Issue #45, opened 2014-11-11
service definition file will be changed frequently


hco:
Hi,

for some reason my service defintion files will be changed every few runs, even though nothing changes.

The following file sometimes looks like this:

``` json
{
  "service": {
    "port": 80,
    "name": "clientinterface",
    "id": "clientinterface",
    "tags": [

    ]
  }
}
```

and sometimes like this:

``` json
{
  "service": {
    "tags": [

    ],
    "port": 80,
    "name": "clientinterface",
    "id": "clientinterface"
  }
}
```

This will result in consul restarting every few runs :-(

PR #48, opened 2014-11-17
Consul reload

jsok:
Adds a `consul::reload` resource and notifies it when any config file changes, rather than restarting the whole service.

Should alleviate the side-effects of #45.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -50,5 +50,5 @@ define consul::check(
   File[$consul::config_dir] ->
   file { "${consul::config_dir}/check_${id}.json":
     content => template('consul/check.json.erb'),
-  } ~> Class['consul::run_service']
+  }
 }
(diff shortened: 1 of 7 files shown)


solarkennedy:
I'm pretty sure that 
https://github.com/solarkennedy/puppet-consul/pull/48/files#diff-60ae41fd0a31977447947f59940ee9a4L84

Will make it so that *any* config change will still initiate a restart?

jsok:
Correct, full service restart on any config.json change. As @EvanKrall pointed out above, there's only a handful of config options which can be applied without a restart.
However, a reload will be triggered on any service/check/watch change.

solarkennedy:
But https://github.com/solarkennedy/puppet-consul/pull/48/files#diff-60ae41fd0a31977447947f59940ee9a4L84 operates on the entire config class.

So when a service changes, the service will reload *and* restart.

I think to do this right you have to turn the ~> into a ->, and make sure the config file directly notifies the run_service.

aj-jester:
Closing in favor of https://github.com/solarkennedy/puppet-consul/pull/152

PR #50, opened 2014-11-19
Added ability to have services and watches configured via hiera

split3:
We use hiera extensively for configuration so it would be nice to have this functionality available.


solarkennedy:
Closed by #72

PR #60, opened 2015-01-05
Allow hash keys to be set to undef

bodepd:
This patch protects against the case where
hash values happen to be set to :undef (Puppet's
nilish representation). Previously, having nil
values set to null caused a stack trace, now,
with this patch, those hash values are simply
dropped.

It also gets rid of some trailing white-spaces


solarkennedy:
Great! Can you add a test to ensure this works fully and will continue to work?

Something like...
```ruby
context "handles undef correctly" do
  let(:params) => {{
    config_hash => {
       something => 'undef'? (maybe :undef ?)
    }
  }}
  it { should compile }
  it { should_not contain_file('consul.json').with_content(/: undef/)? }
end
```

solarkennedy:
Nice!

PR #62, opened 2015-01-10
Add class to secure local ports

bodepd:
While consul does some things great,
by default, it introduces some fairly
severe security risks.

Some of its functionality allows commands
to be run from other machines (exec, service
checks). Since these commands are run behind
services bound to a port on localhost, it means
that by default, any user on a system registered with
consul can execute code on any other registered system.

This patch adds a class that configures firewall rules
to limit access to only root and consul user to interact
with the API.

Unfortunatly, the uid feature for the firewall module does
not work as documented. I am working on the following
upstream patch to resolve the issue:

  https://github.com/puppetlabs/puppetlabs-firewall/pull/466

It probably makes sense to hold off merging this until that
upstream patch is merged and released, but I wanted to go ahead
and put this patch out there becuase:

1. as a reference point, anyone running consul in prod should be
adding these rules
2. to start a conversation to ensure the patch is acceptable as-is
while I work with upstream on the other patch.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,5 +2,6 @@ fixtures:
   repositories:
     stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
     staging: "https://github.com/nanliu/puppet-staging.git"
+    firewall: "git://github.com/puppetlabs/puppetlabs-firewall"
   symlinks:
     consul: "#{source_dir}"
--- a/Modulefile
+++ b/Modulefile
@@ -9,3 +9,4 @@ project_page  'https://github.com/solarkennedy/puppet-consul'
 
 dependency 'puppetlabs/stdlib', '>= 0.1.6'
 dependency 'nanliu/staging', '>=0.4.0'
+dependency 'puppetlabs/firewall', '>=1.3.0'
--- a/manifests/secure_local_ports.pp
+++ b/manifests/secure_local_ports.pp
@@ -1,0 +1,27 @@
+#
+# Secures consul usage by ensuring that only the
+# specified users can access internal consul services
+# by default
+#
+class consul::secure_local_ports(
+  $internal_ports = ['8300', '8400', '8500'],
+  $consul_user    = 'consul'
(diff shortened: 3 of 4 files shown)


solarkennedy:
Sorry, it is not inside the scope of this module to include firewall rules.

I am a big believer in the role/profile/module pattern of puppet developement, this would go in something like `profile_consul` which might include firewall rules, logrotate, monitoring, sudoers, etc.

I would include this code in some sort of extra file like EXAMPLES.md ?

bodepd:
no problem, I was trying to share outside of my profile b/c it's slightly scary to run consul without these rules and the security risks may not be entirely obvious to people getting started.

solarkennedy:
I agree. I wish there were more "profile_bla" puppet modules out there the wild, but too often they contain too much business logic.

Can you open source your profile_consul?

EvanKrall:
It might be a good idea to set [`disable_remote_exec`](https://www.consul.io/docs/agent/options.html#disable_remote_exec) by default.

@bodepd I don't think I see how services or checks can be used to execute code remotely. I thought those were defined on the local agent only.

bodepd:
3 concerns:

1. http api's that looks like they allow remote command execution

* /v1/catalog/register
* /v1/agent/service/register

I did not verify remote code execution with any feature besides consul exec, so it is possible that I misunderstood the documentation.

2. consul exec (and the ability for arbitrary users to execute code remotely)

3. privilege escalation - allowing any user on a system to be able to execute commands as the consul user (or whatever user consul checks get executed as)

bodepd:
@solarkennedy I'm happy to share, everything I work on is open source. This patch was take from a refactor patch that I was working on this week: https://github.com/JioCloud/puppet-rjil/pull/290. Most of the consul specific bits are in manifests/jiocloud/consul

solarkennedy:
This is good stuff. Thank you for sharing! This is the right level (imho) to do this work, where you combine the consul, firewall, dnsmasq and ssl stuff together.

PR #63, opened 2015-01-14
Better OS support for init_style

avishai-ish-shalom:
Support non LTS ubuntu versions, etc.

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -234,7 +234,6 @@ describe 'consul' do
   context "On a redhat 6 based OS" do
     let(:facts) {{
       :operatingsystem => 'CentOS',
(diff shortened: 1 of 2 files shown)


spec/classes/init_spec.rb
```diff
@@ -255,6 +257,7 @@
     let(:facts) {{
       :operatingsystem => 'Fedora',
       :operatingsystemmajrelease => 20,
```
solarkennedy: is operatingsystemmajrelease no longer used? if so can you remove it to prevent confusion?

avishai-ish-shalom: sure.

avishai-ish-shalom:
fixed

solarkennedy:
Thanks!

PR #64, opened 2015-01-15
Adding "status" to the debian init script

paulhamby:
This adds a status argument to the debian init script. Without the status, I experienced puppet continually trying to start consul, even though it was already running.

Thanks for the great module!


solarkennedy:
Thanks!

PR #67, opened 2015-01-19
Adding ability to disable managing of the service

sedan07:
Useful for instance when running on a RHEL7 docker container which uses fakesystemd


solarkennedy:
Thanks!

PR #68, opened 2015-01-20
Support for Amazon OS

dcoxall:
Tried using this module on EC2 and it failed to start. Looks as though it is because it didn't detect the O/S and so didn't create an init script.

/cc @telent


nanorepublica:
:+1:

solarkennedy:
Thanks!

PR #69, opened 2015-01-21
Update install.pp

ianlunam:
For consistancy and to avoid warnings


solarkennedy:
It looks like there are a lot of things we are not quoting in there. Can you do a PR with more `bundle exec rake lint` fixes?

PR #66, opened 2015-01-16
Some linting fixes and resolves joining wan not actually joining the wan

justicel:

PR #70, opened 2015-01-21
One bug fixe and one depreciation fix

chrisleavoy:


solarkennedy:
The wan fix is a dupe of #66.

What is deprecated about the sudo method? I suppose su would work just as well but it looks a little messy?

chrisleavoy:
Yeah, I agree its less clean but it seemed like the best approach for compatibility. Maybe its less common than I thought, but I've run into the security practice of sudo not being permitted without a terminal (tty). Default is allow in 14.04. I thought I'd submit it upstream so I wouldn't have to maintain my own fork.  I see #66 is merged now. So maybe when i get some free time I can resubmit this without breaking the test?

solarkennedy:
Yes please :)

I agree that sudo is weird.

split3:
Closing this due to it is over a year old.  If this is still needed please re-open and fix the failing tests.

PR #71, opened 2015-01-22
Fix Puppet 3.7.3 giving evaluation error in run_service.pp

tayzlor:
If I run this module out on a CentOS 6.5 box with Puppet 3.7.3 using future parser I get the following error - 

Error: Evaluation Error: Error while evaluating a Resource Statement, Illegal fully qualified name in file /etc/puppet/modules-extra/consul/manifests/run_service.pp at line 32

This looks like some scoping issue because it cannot lookup the variable $_config_hash. This seems to be down to the fact it starts with an underscore. If I change the variable to use $config_hash_real (no underscore) it seems to work out just fine.

This might actually be down to some bug in Puppet itself, I couldn't see anything in the docs (for Puppet) about variable names and scoping, apart from the fact you can no longer have variables that start with a capital letter.

This should circumvent the issue, but might be worth raising an issue on the Puppet project separately too as this seems a bit weird to me?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,15 +69,15 @@ class consul (
   validate_hash($config_hash)
   validate_hash($config_defaults)
 
-  $_config_hash = merge($config_defaults, $config_hash)
-  validate_hash($_config_hash)
+  $config_hash_real = merge($config_defaults, $config_hash)
+  validate_hash($config_hash_real)
 
-  if $_config_hash['data_dir'] {
-    $data_dir = $_config_hash['data_dir']
+  if $config_hash_real['data_dir'] {
+    $data_dir = $config_hash_real['data_dir']
   }
 
-  if $_config_hash['ui_dir'] {
-    $ui_dir = $_config_hash['ui_dir']
+  if $config_hash_real['ui_dir'] {
+    $ui_dir = $config_hash_real['ui_dir']
   }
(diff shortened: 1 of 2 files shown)


solarkennedy:
The wan thing is covered in #66, can you rebase with only the underscore change?

tayzlor:
#66 didn't yet get merged to master. Are you planning to merge that or do you want me to rebase against that fork? in the line that joins the WAN it references the underscore variable, so that would need updated.

solarkennedy:
Yes. #66 is merged now. Please rebase.

What do you think about adding the future parser to the rspec stuff so we can (potentially) catch other bugs?

https://github.com/rodjek/rspec-puppet/issues/182#issuecomment-46182054

tayzlor:
Thanks, Rebased!
Yeah might be worth creating a separate issue to add in the future parser stuff.

solarkennedy:
Thank you for rebasing, I know it is a pain. :)

99% of the time the future parser shouldn't be an issue, but I guess this was one of those times?

Seems like modules shouldn't have to test against the future parser unless they explicitly are using it?

tayzlor:
Yeah it's all a bit weird, would be much easier if Puppet just settled on a parser (move to the future one!) and not have this flip/flopping between both.

PR #72, opened 2015-01-22
Add support for providing watches/checks/services via hiera 

tayzlor:
This related to https://github.com/solarkennedy/puppet-consul/pull/50

I fixed up the tests and added support for providing watches/services/checks via hiera args to the class


solarkennedy:
Perfect!! This is exactly what I had in mind.
Thank you very much for writing this!

Issue #74, opened 2015-01-23
Consul init scripts sometimes not installed in the correct order


tayzlor:
Because we install the init scripts in install.pp, what can happen is that sometimes the init script will get put in place and then then package will be installed over the top, blowing the init script from the module away. If the package that was installed has a slightly different init script, this could lead to the service running with unexpected circumstances (as I have experienced). This is because there is no dependency between the installation > the placing of the init script.

To solve this the easiest fix would be to just move the logic from install.pp into either config.pp or run_service.pp. This would ensure the dependency works out correctly because of the ordering in init.pp

PR #75, opened 2015-01-23
Move init script to config.pp to ensure it gets set AFTER the package gets installed

tayzlor:
This addresses #74


solarkennedy:
cool

PR #84, opened 2015-02-23
Updated the params for OracleLinux Support

te206676:
Added a small change to the Redhat Operating systems group for support of OracleLinux.


solarkennedy:
Cool. Can you update metadata.json as well?

te206676:
Updated the metadata.json to include Oracle Linux as a supported OS.

solarkennedy:
Thank you!

PR #86, opened 2015-02-25
Expose the ID parameter for service definitions

codesplicer:
Its useful to be able to override the ID parameter, as these need to be unique per node.


solarkennedy:
Can you add a docstring for this new parameter?

codesplicer:
I closed this PR because I pushed to master in my fork instead of branches. Will reopen as a new PR

PR #87, opened 2015-02-25
sysv & debian init updates to kill or leave

### LINKED ISSUE #85: consul should not 'leave' for init script 'stop' action
When consul is provided with the `leave` subcommand, the node is removed from the cluster. This requires that the cluster be added back in at reboot with a `join` action. This breaks expected behavior for the service, in which the node automatically rejoins the cluster upon service start. A more acceptable init-style script behavior would be to kill the consul process so it does not remove itself from the cluster.

runswithd6s:
This fixes #85. The 'stop' action in init scripts for sysv and Debian
will only 'leave' the cluster if acting as an agent. When running as a
server, as determined by a call to `consul info`, kill the process
instead.

Both updates also enforce the use of a PID file located in a
/var/run/consul directory, writeable by the consul::user configured in
Puppet.


runswithd6s:
I haven't tested this in AWS or Debian. Please test. Works great in Centos 6.6.

solarkennedy:
Thanks!

PR #88, opened 2015-02-26
Expose id parameter for service definitions

codesplicer:
Its useful to be able to override the service ID as these must be unique per node.


solarkennedy:
Thanks!

PR #89, opened 2015-02-26
Consul expects prefix rather than keyprefix in watch config

codesplicer:
Consul expects the 'prefix' parameter rather than 'keyprefix' in watch definitions. Test has been updated too.


solarkennedy:
I don't think so?

On type "keyprefix" the key should be "keyprefix" according to the docs:
http://www.consul.io/docs/agent/watches.html

codesplicer:
The 'type' of watch is keyprefix yes, but to specify the actual prefix to watch, the parameter is 'prefix', if you scroll further down to the example config. I've also tested this in a vagrant machine running consul 0.5.0 and confirmed as such.

solarkennedy:
Ah! I understand. You are right.

Thank you!

PR #91, opened 2015-03-07
Add consul_acl type and provider

michaeltchapman:
This patch adds support for managing consul ACLs.
Rules can be specified in a ruby hash, while type
must be either 'client' or 'management'.

Example:
consul_acl { 'ctoken':
  ensure => 'present',
  rules  => {'key' => {'test' => {'policy' => 'read'}}},
  type   => 'client',
}

While consul does not enforce unique names, the provider will
not behave consistently if the user manually creates ACLs with the
same name and subsequently attempts to manage them with puppet.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -119,6 +119,19 @@ See the check.pp docstrings for all available inputs.
 You can also use `consul::checks` which accepts a hash of checks, and makes
 it easy to declare in hiera.
 
+## ACL Definitions
+
+```puppet
+consul_acl { 'ctoken':
+  ensure => 'present',
+  rules  => {'key' => {'test' => {'policy' => 'read'}}},
+  type   => 'client',
+}
+```
+
+Do not use duplicate names, and remember that the ACL ID (a read-only property for this type)
+is used as the token for requests, not the name
+
 ##Limitations
 
 Depends on the JSON gem, or a modern ruby.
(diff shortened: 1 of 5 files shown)


solarkennedy:
Sweet! 

I don't know much about developing types/providers. Does it play well when bootstrapping and the consul server doesn't exist yet?

There are no tests.. so I can't really tell if it works or not, even syntax-wise. I don't know what it takes to test a type/provider. Can you provide a basic test?

michaeltchapman:
The short answer is that during bootstrapping consul won't accept writes, so if you plan on having 3 servers and boot the first one while defining ACLs, the ACL resources will error. This is correct behaviour imo. I probably need to amend the prefetch/instances methods to handle that case more gracefully (right now I think it will throw a confusing error), and I also realised I need to define an autorequire for the consul service.

As far as tests go, writing tests for providers appears to be more complex than writing the provider itself, so...maybe? :)

michaeltchapman:
Add spec tests for custom type.

solarkennedy:
Nice. Thank you for writing tests. This will give future edits more confidence in this code.

Also thank you for the acl entry point in init.pp to make it easier to deploy with hiera.

Hopefully this will help make everyone's puppet-deployed consul installs *able* to be more secure!

PR #92, opened 2015-03-08
make module work with future parser

duritong:
versioncmp can't be used with future parser as it currently is being used, while another usage doesn't work with the current parser. :-/ So I refactored things, it's a bit more verbose, but works on both parsers.


solarkennedy:
Thanks!

PR #93, opened 2015-03-12
Use fixed titles for packages so that dependencies are easier

bodepd:
I'm submitting this patch on behalf of @sorenh, b/c we are carrying it downstream at the moment. Having static titles makes it possible for us to set up dependencies on resources inside of the classes without having to know the values of variables. I think it's a good practice in general to only let namevars be dynamic and always make title's static.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -51,13 +51,15 @@ class consul::install {
 
   } elsif $consul::install_method == 'package' {
 
-    package { $consul::package_name:
+    package { 'consul':
       ensure => $consul::package_ensure,
+      name   => $consul::package_name,
     }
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
I'm down with this, however you might be better served by depending on the subclasses, like
`````
something { 
  require => Class['consul::install']
}
`````
instead of peaking into the exact package name?

But I would still accept this, however the tests are failing on lint.

bodepd:
I'll update the lint failures.
As to class level dependencies, I just find them to be super clunky in Puppet. They can lead to dependency cycles in some cases and the class containment issues means that you can never trust class level dependencies to work without reading the code.

PR #94, opened 2015-03-13
Provided very basic compatibility with 0.5.0

centromere:


solarkennedy:
Cool. I like your use of delete_undef_values, if it works?

I'm a little skeptical about your removal of all the validation logic?

Also tests are failing:
https://travis-ci.org/solarkennedy/puppet-consul/jobs/54294185

We have a lot of tests to ensure that the json that comes out of this module is *correct*, and especially correctly converts "ints" to real ints in the json. Please work on the tests.

Also I think "id" is a required input to the json?

centromere:
The config files are validated by consul itself, and as such I don't see a reason to duplicate the logic.

I will look closer at the tests.

As per https://consul.io/docs/agent/services.html:
> A service definition must include a name and may optionally provide an id, tags, address, port, and check. The id is set to the name if not provided.

And https://consul.io/docs/agent/checks.html:
> Each type of definition must include a name and may optionally provide an id and notes field. The id is set to the name if not provided.

solarkennedy:
One reason for the validation in puppet is to prevent a human mistake of adding a new check that rolls out to the entire consul cluster, puppet restarts it, doesn't come back because of a syntax error or something, and now your whole cluster is down?

But I agree there is duplication and it can be a fine-line, but I'm a fan of fast-feedback.

thijsterlouw:
Is it possible to get this merged? We need to use 0.5.0 of Consul and would really like to use this puppet module as-is

solarkennedy:
I don't mind the removed validation logic as much as the failing tests?

PR #95, opened 2015-03-26
Notify consul service

dominikschulz:
This PR changes the notification dependency to notify the Consul Service instead of the Run Service Class.

In my tests this did result in correctly restarted consul instances after configuration changes. Without this patch consul was lacking necessary restarts/reloads.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -71,5 +71,5 @@ define consul::check(
   File[$consul::config_dir] ->
   file { "${consul::config_dir}/check_${id}.json":
     content => template('consul/check.json.erb'),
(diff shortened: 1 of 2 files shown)


solarkennedy:
This is surprising and shouldn't be necessary.
The reason for notifying the Class is so that it can be controlled via $manage_service, etc.

solarkennedy:
Can you give me some output where you make a config change and puppet does *not* restart consul?

solarkennedy:
Closing due to inactivity. @dominikschulz recently this code has changed to try to minimize the restarts and does reloads when possible. Can you try again on master and open a new issue if it still happens?

PR #96, opened 2015-03-26
Allow multiple versions to be installed

jsantiago:
Using this patch downstream to make upgrading a little easier.

* Added ability to specify download directory
  * Defaults to current location of `/usr/local/bin`
* Downloads into versioned directory
* Creates symlink to versioned binary

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -90,6 +90,10 @@ class consul (
     warning('data_dir must be set to install consul web ui')
   }
 
+  if $config_hash_real['download_dir'] {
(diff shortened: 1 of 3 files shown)


solarkennedy:
This seems a little messy. Puppet would be making new folders in /usr/local/bin? How would they get cleaned up?

The whole point of the staging module is so we *don't* have to deal with directories with binaries in them?

jsantiago:
Fair points. In our downstream version we seem to have lost the staging module functionality (!) so it would make more sense to correct that.

Thanks for your feedback!

PR #101, opened 2015-04-04
Tweaks to the sysv file

jaxxstorm:
A few minor tweaks to the sysv startup file which give a nicer output when starting and stopping

### DIFF
--- a/templates/consul.sysv.erb
+++ b/templates/consul.sysv.erb
@@ -53,7 +53,7 @@ start() {
         [ -f $PID_FILE ] && rm $PID_FILE
         daemon --user=<%= scope.lookupvar('consul::user') %> \
             --pidfile=$PID_FILE \
-            { $CONSUL agent -pid-file ${PID_FILE} -config-dir $CONFIG <%= scope.lookupvar('consul::extra_options') %> &>> $LOG_FILE & }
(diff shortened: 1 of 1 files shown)


templates/consul.sysv.erb
```diff
@@ -52,9 +52,10 @@ start() {
         mkrundir
         [ -f $PID_FILE ] && rm $PID_FILE
         daemon --user=<%= scope.lookupvar('consul::user') %> \
-            --pidfile="$PID_FILE" \
-            "$CONSUL" agent -pid-file "${PID_FILE}" -config-dir "$CONFIG" <%= scope.lookupvar('consul::extra_options') %> >> "$LOG_FILE" &
+            --pidfile=$PID_FILE \
+            " { ${CONSUL} agent -pid-file ${PID_FILE} -config-dir ${CONFIG} <%= scope.lookupvar('consul::extra_options') %> &>> ${LOG_FILE} & } ;  echo \$! >| ${PID_FILE}"
         retcode=$?
+        echo
```
solarkennedy: On :51 we do the echo -n. Here would be the place to say "done"? or "started"?

templates/consul.sysv.erb
```diff
@@ -68,9 +69,10 @@ stop() {
 
         # If acting as a server, or if leave failed, kill it.
         mkpidfile
-        killproc $KILLPROC_OPT $CONSUL -9
+        killproc $KILLPROC_OPT $CONSUL -INT
 
         retcode=$?
+        echo
```
solarkennedy: On :64 we do the echo -n. Here would be the place to say "killed"?

Issue #107, opened 2015-04-14
umask feature breaks CentOS init scripts


split3:
The newly added umask feature causes failures on startup with the following error:

`/etc/init.d/consul: Usage: daemon [+/-nicelevel] {program} [FAILED]`

This happens with CentOS 6.5 and 6.6

split3:
Should also note, that a nice side effect of this is that it brought down my entire cluster.

split3:
Closing since this change was reverted.

PR #108, opened 2015-04-14
Revert "Allow setting of the umask for the consul daemon."

sjoeboo:
Reverts solarkennedy/puppet-consul#106


EvanKrall:
Seems like a good idea until I fix #107.

PR #111, opened 2015-04-20
Update puppetlabs-stdlib dependency to 4.2.0 for the delete_undef_values function

### LINKED ISSUE #109: delete_undef_values required stdlib 4.2.0, dependency not set properly
Hey guys, first of all: I appreciate all the hard work you're doing here, this module is great.

I noticed a small problem when pulling from master (specifically I'm working off of: 0bbbea631780b284d36dff955716adab8c0652e0). When pulling the module via librarian-puppet it checks the dependencies in the [metadata.json](metadata.json) file. I currently have puppetlabs-stdlib 4.1.0 installed, which passed the dependency check in the metadata.json file, because that file claims your module only requires puppetlabs-stdlib version 0.1.6 or greater. However, a the new function delete_undef_values introduced in [this commit](cce035cbc59df5a410948e287f3637870b819bec) requires at least puppetlabs-stdlib 4.2.0.

It seems that the metadata.json file needs to be updated to reflect that fact. I can submit a PR if you desire, I'm more than willing to do that.

Thanks!

zxjinn:
I also perused through the rest of the manifests looking for other functions that might be in a version later than 4.2 and didn't see anything else.
Closes #109

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -9,7 +9,7 @@
   "issues_url": "https://github.com/solarkennedy/puppet-consul/issues",
   "description": "Configures Consul by Hashicorp",
   "dependencies": [
-    {"name":"puppetlabs/stdlib","version_requirement":">= 0.1.6 <5.0.0"},
+    {"name":"puppetlabs/stdlib","version_requirement":">= 4.2.0 <5.0.0"},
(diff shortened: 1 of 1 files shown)


solarkennedy:
Super. Explicit is good.

PR #112, opened 2015-04-20
Update README, ensure passingonly is a bool

### LINKED ISSUE #97: passingonly needs to be a boolean for watch type
## Problem

The documentation contains the example:

```puppet
consul::watch { 'my_watch':
  ...
  passingonly => 'true',
}
```

This results in a config file like:

```json
{
  "watches": [
    {
     ...
     "passingonly": "true"
    }
  ]
}
```

This will in turn cause an issue for Consul starting with an error similar to:

```
==> Failed to parse watch (map[string]interface {}{"passingonly":"true", "handler":"/usr/bin/runpuppet.sh"}): Expecting %!s(MISSING) to be a boolean
```

Changing the config file by hand to the following fixes the problem.

```json
"passingonly": true
```

Passing in the value to Puppet as a boolean also fixes the problem.

```puppet
consul::watch { 'my_watch':
  ...
  passingonly => true,
}
```

## Fix

* Docs could be updated to show working values
* Manifests could validate values being passed are booleans

## Additional thoughts

* This may affect other types which take booleans, and currently show passing strings in the READM.
* This might be version specific, just in case:

```
consul --version
Consul v0.4.1
Consul Protocol: 2 (Understands back to: 1)
```

### LINKED ISSUE #110: README for consul::service is out of date
In the README the `Service Definition` section code example looks like it's showing an older way of configuration. The redis example should look like so:
```puppet
  ::consul::service { 'redis':
    checks  => [
      { 
        script   => "/usr/local/bin/check_redis.py",
        interval => "10s"
      }
    ],
    port    => 8000,
    tags    => ['master']
  }
```

Just like my other issue I just made, I'm more than willing to submit a PR if you don't want to do the legwork.

zxjinn:
There are two commits:
First-
Add top-level namespaces for everything (prefix ::) for future compatibility.
Differentiate agent and server configs, provide examples for both.
Update consul::service code example to match some recent manifest changes to that file.
Alphabetize parameters.
Change 'true' to '/bin/true' to prevent confusion about what 'true' means in that context.
Delete trailing whitespace
Closes #110, #97

I tested all of the code in the README except `consul_acl`, I couldn't get it to work on my system.. kept timing out. The Puppet/Ruby code was fine, as far as I could tell, but I couldn't ensure it did what it was supposed to.

Second-
Ensure passingonly is a bool
Closes #97

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -22,33 +22,49 @@ on the 4 series. Do *not* pull from master.
 
 ##Usage
 
+To set up a single consul server, with several agents attached:
+On the server:
 ```puppet
-class { 'consul':
+class { '::consul':
   config_hash => {
-      'datacenter' => 'east-aws',
-      'data_dir'   => '/opt/consul',
-      'log_level'  => 'INFO',
-      'node_name'  => 'foobar',
-      'server'     => true
+    'bootstrap_expect' => 1,
+    'data_dir'         => '/opt/consul',
+    'datacenter'       => 'east-aws',
+    'log_level'        => 'INFO',
+    'node_name'        => 'server',
+    'server'           => true,
+  }
+}
+```
+On the agent(s):
+```puppet
+class { '::consul':
+  config_hash => {
+    'data_dir'   => '/opt/consul',
+    'datacenter' => 'east-aws',
+    'log_level'  => 'INFO',
+    'node_name'  => 'agent',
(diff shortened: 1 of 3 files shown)


zxjinn:
ack! a failure! I'll fix it tomorrow. Looks like the test is expecting a quoted string, too!

solarkennedy:
Thanks! Super great.

PR #114, opened 2015-05-05
Upgrade/Downgrade url method

aj-jester:
This allows us to upgrade/downgrade consul's if the installation method is URL.

### DIFF
--- a/lib/facter/consul.rb
+++ b/lib/facter/consul.rb
@@ -1,0 +1,24 @@
+require 'facter'
+
+# Example:
+# $ consul -v
+# Consul v0.5.0
+# Consul Protocol: 2 (Understands back to: 1)
+consul_v = Facter::Util::Resolution.exec("consul -v")
+
+# If consul is not installed, the command above will
+# fail and return nil
+unless consul_v.nil?
+  
+  # Split the string into an array with 2 elements
+  # and do some ruby shenanigans to extract consul
+  # version and protocol
+  consul_v = consul_v.split("\n")
+  consul_version = consul_v.first.split.last.gsub('v', '')
+  consul_protocol = consul_v.last.split[2]
+  
+  # Finally add the values to facter
+  Facter.add(:consul_version) { setcode { consul_version } }
+  Facter.add(:consul_protocol) { setcode { consul_protocol } }
+end
+
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,13 +1,15 @@
 # == Class consul::intall
 #
+# Installs consule based in the parameters from init
+#
 class consul::install {
 
   if $consul::data_dir {
-    file { "${consul::data_dir}":
+    file { $consul::data_dir:
       ensure => 'directory',
-      owner => $consul::user,
-      group => $consul::group,
-      mode  => '0755',
+      owner  => $consul::user,
+      group  => $consul::group,
+      mode   => '0755',
     }
   }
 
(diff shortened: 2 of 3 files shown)


lib/facter/consul.rb
```diff
@@ -0,0 +1,24 @@
+require 'facter'
+
+# Example:
+# $ consul -v
+# Consul v0.5.0
+# Consul Protocol: 2 (Understands back to: 1)
+consul_v = Facter::Util::Resolution.exec("consul -v")
```
solarkennedy: This seems like it could be potentially brittle or hard to refactor.

Can you add a test for this to stub out exec with the example that your provided in the comments and assert the desired expected output?

lib/facter/consul.rb
```diff
@@ -0,0 +1,24 @@
+require 'facter'
+
```
solarkennedy: Seems like this should be "confined" on something. I guess not really because it could be windows/mac/linux

lib/facter/consul.rb
```diff
@@ -0,0 +1,24 @@
...
+  
+  # Split the string into an array with 2 elements
+  # and do some ruby shenanigans to extract consul
+  # version and protocol
+  consul_v = consul_v.split("\n")
+  consul_version = consul_v.first.split.last.gsub('v', '')
+  consul_protocol = consul_v.last.split[2]
+  
+  # Finally add the values to facter
+  Facter.add(:consul_version) { setcode { consul_version } }
```
solarkennedy: Adding the Facter.add here is fine, but it would be better (I think) if the secode was at a higher level.
Mostly so you can wrap this all in a timeout. All facts should timeout.

manifests/install.pp
```diff
@@ -18,17 +18,43 @@
...
+      file { "${consul::bin_dir}/consul":
+        owner => 'root',
+        group => 0, # 0 instead of root because OS X uses "wheel".
+        mode  => '0555',
+      }
+
+    # Upgrading existing installation of consul
+    } elsif ($::consul_version != $consul::version) {
+
+      exec { "removing consul v${::consul_version}":
```
solarkennedy: Am I reading this right that the only difference between the fresh and non-fresh install is this exec?

If that is the case, can you wrap only the exec in the if block? I think that would be clearer and reduce code duplication.

manifests/install.pp
```diff
@@ -18,17 +18,43 @@
...
+        target => $consul::bin_dir,
+      } ->
+      file { "${consul::bin_dir}/consul":
+        owner => 'root',
+        group => 0, # 0 instead of root because OS X uses "wheel".
+        mode  => '0555',
+      }
+
+    # Upgrading existing installation of consul
+    } elsif ($::consul_version != $consul::version) {
```
solarkennedy: This seems like an iffy way to do this. Puppet works better in a kind of "tell don't ask" kind of way.

Can you think of a better way to do this?

Heck what if we just did this?

```puppet
file { '/etc/consul_version': content => $consul::version } ~>
exec { "rm -rf bin/consul": refreshonly => true  }
```

manifests/install.pp
aj-jester: Also the service needs to be restarted (line 56)

lib/facter/consul.rb
aj-jester: Def, I can add a test for this.

Also, regarding refactoring, do you think its better to inject those values in a hash or something? So incase in the future we need more than just `version` or `protocol` it would be easily manageable? Or did you mean some other way to refactor? I can imagine adding more consul facts in the future.

solarkennedy: I think just normal strings is ok.

I'm mostly concerned about guarding against consul not being available, string parsing, and timeouts. I've been bit by all those things.

There is another thing I would like to suggest in this refactor and that is to make it so it returns "unknown" when not available. This makes the fact easier to use in contexts when strict variables is enabled in puppet.

We do not currently test with strict variables on, but I think we should.

manifests/install.pp
solarkennedy: Thats fine, although I would rather have the entire install class notify the entire run_service class per this pattern
https://www.devco.net/archives/2012/12/13/simple-puppet-module-structure-redux.php

PR #118, opened 2015-05-18
Upgrade and Downgrade with url

thijsterlouw:
This allows users to upgrade (and downgrade) their consul version, even when using the 'url' install_method.

It also notifies the service, so it gets properly restarted.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -15,37 +15,50 @@ class consul::install {
 
   if $consul::install_method == 'url' {
 
+    $version = $consul::version
+    $consul_archive_target = "consul_${version}.zip"
+    $consul_web_ui_archive_target = "consul_web_ui_${version}.zip"
+
     if $::operatingsystem != 'darwin' {
       ensure_packages(['unzip'])
     }
-    staging::file { 'consul.zip':
+    staging::file { $consul_archive_target:
       source => $consul::download_url
     } ->
-    staging::extract { 'consul.zip':
-      target  => $consul::bin_dir,
-      creates => "${consul::bin_dir}/consul",
+    file { "${consul::bin_dir}/consul_${version}/":
+      ensure => 'directory',
+      owner  => 'root',
+      group  => 0,
+      mode   => '0755',
+    } ->
+    staging::extract { $consul_archive_target:
+      target  => "${consul::bin_dir}/consul_${version}/",
+      creates => "${consul::bin_dir}/consul_${version}/consul",
     } ->
     file { "${consul::bin_dir}/consul":
-      owner => 'root',
-      group => 0, # 0 instead of root because OS X uses "wheel".
-      mode  => '0555',
+      owner  => 'root',
+      group  => 0, # 0 instead of root because OS X uses "wheel".
+      mode   => '0555',
+      target => "${consul::bin_dir}/consul_${version}/consul",
+      notify => Class['consul::run_service'],
     }
 
+
     if ($consul::ui_dir and $consul::data_dir) {
-      file { "${consul::data_dir}/${consul::version}_web_ui":
+      file { "${consul::data_dir}/${version}_web_ui":
         ensure => 'directory',
         owner  => 'root',
         group  => 0, # 0 instead of root because OS X uses "wheel".
         mode   => '0755',
       } ->
-      staging::deploy { 'consul_web_ui.zip':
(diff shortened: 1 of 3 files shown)


thijsterlouw:
Very similar to https://github.com/solarkennedy/puppet-consul/pull/114 , but I think this change is simpler

manifests/params.pp
```diff
@@ -10,7 +10,7 @@
   $package_ensure    = 'latest'
   $ui_package_name   = 'consul_ui'
   $ui_package_ensure = 'latest'
-  $version = '0.5.0'
+  $version = hiera('consul::version', '0.5.0')  # ensure defaults resolve from hiera as well
```
xakraz: It is not advised to call Hiera from a Module ...

Best practices are to use the Role / Profile pattern to compose nodes definition and make Hiera calls from these classes which are business specific anyway .

manifests/install.pp
```diff
@@ -15,37 +15,50 @@
...
     } ->
     file { "${consul::bin_dir}/consul":
-      owner => 'root',
-      group => 0, # 0 instead of root because OS X uses "wheel".
-      mode  => '0555',
+      owner  => 'root',
+      group  => 0, # 0 instead of root because OS X uses "wheel".
+      mode   => '0555',
+      target => "${consul::bin_dir}/consul_${version}/consul",
+      notify => Service['consul'],
```
aj-jester: This will break puppet if `$consul::manage_service` is `false`. I would suggest notifying the class, like so: `Class['consul::run_service']`

manifests/params.pp
thijsterlouw: In our puppet setup we have various roles that use the Consul module. We don't want to write the Consul version in each of the roles (or their hiera counterpart). The old code forces the user of the Consul module to explicitly pass the version for Consul to the module in order to create the correct download url, whereas the actual version in the init.pp would be set from hiera. This seems highly counter-intuitive to me. We don't have the concept of Profiles.

manifests/install.pp
thijsterlouw: Good point, so I assume 
-      notify => Service['consul'],
+     notify => Class['consul::run_service']

aj-jester: yup :+1:

manifests/params.pp
solarkennedy: I'm also going to veto an explicit heira lookup here.

I don't understand why `$consul::version` in hiera would not be picked up using the native hiera bindings?

It may be counter-intuitive to you, but this is the documented way that variable precedence takes place:
https://docs.puppetlabs.com/hiera/1/puppet.html#automatic-parameter-lookup

thijsterlouw: ```$consul::version``` works fine
but in init.pp we have defaults:
```
$download_url = "https://dl.bintray.com/mitchellh/consul/${version}_${os}_${arch}.zip",
$ui_download_url = "https://dl.bintray.com/mitchellh/consul/${version}_web_ui.zip",
```
These defaults apparently do *not* get populated by hiera. Eg if you have in hiera:
```consul::version: 0.5.1``` and in params.pp ```$version = '0.5.0'``` this means your url will become ```https://dl.bintray.com/mitchellh/consul/0.5.0_web_ui.zip``` instead of 0.5.1 (which is used in other places because init.pp:version is correctly resolved from hiera)

Comment (aj-jester) [inline manifests/params.pp]: Something must be weird with your hiera setup then.

We use hiera exactly like you mention:

Eg in hiera
```
consul::version: 0.5.1
```

and in params.pp
```
$version = '0.5.0'
```

and the download url becomes:
```
https://dl.bintray.com/mitchellh/consul/0.5.1_web_ui.zip
```

thijsterlouw: > Oh, do you mean those urls as params in init.pp?
Not sure what you mean, but end-result is that if I only set consul::version in hiera, my download comes from the version-number as set in params. I'll test it again!

aj-jester: Are we still blocked on this? Can we keep this the way it was (without hiera) and merge it?

aj-jester: @thijsterlouw I think the problem you are experiencing with hiera is fixed here https://github.com/solarkennedy/puppet-consul/issues/129

split3:
This functionality should be handled now by #202

PR #125, opened 2015-05-26
Set provider on consul service

albustax:
Hi,

If provider is not set, service management does not work on Debian 8, we have an error:

`Error: /Stage[main]/Consul::Run_service/Service[consul]: Could not evaluate: Could not find init script for 'consul' `

This PR set provider defined in `consul::params`

### DIFF
--- a/manifests/run_service.pp
+++ b/manifests/run_service.pp
@@ -12,9 +12,10 @@ class consul::run_service {
 
   if $consul::manage_service == true {
     service { 'consul':
-      ensure => $consul::service_ensure,
-      name   => $init_selector,
-      enable => $consul::service_enable,
+      ensure   => $consul::service_ensure,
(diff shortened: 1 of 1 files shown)


solarkennedy:
I'm hesitant, usually puppet figures out the provider correctly, we shouldn't have to be explicit here.

can you run `puppet resource service consul --debug` and pastebin?

albustax:
http://pastebin.com/i98crtUM

solarkennedy:
Hmm, I'm pretty new to systemd, but is it more that the service name is simply "consul.service"? Per your pb there was a `Service[consul.service]` in there.

solarkennedy:
Sorry for the late review. This seems like a good thing, and init systems are hard, especially when there can be multiple providers per os. Thanks!

PR #127, opened 2015-05-28
Add the ability to deploy a check script

albustax:


manifests/check.pp
```diff
@@ -32,22 +42,35 @@
...
+  $interval      = undef,
+  $service_id    = undef,
+  $timeout       = undef,
+  $notes         = undef,
 ) {
   include consul
 
+  # Deploy script if necessary
+  if $script_source {
+    file { "${id}":
```
solarkennedy: shouldn't this just be $script and then we don't need the $script_path variable?

manifests/check.pp
```diff
@@ -32,22 +42,35 @@
...
+      ensure => $check_ensure,
+      source => $script_source,
+      path   => $script_path,
+      mode   => 755,
```
solarkennedy: '0555'

manifests/check.pp
```diff
@@ -32,22 +42,35 @@
...
-  $id         = $title,
-  $ttl        = undef,
-  $http       = undef,
-  $script     = undef,
-  $interval   = undef,
-  $service_id = undef,
-  $timeout    = undef,
-  $notes      = undef,
+  $id            = $title,
+  $check_ensure  = present,
```
solarkennedy: Can this just be called 'ensure' ?

split3:
I'm not sure I'm a fan of having script deployment stuff in this module.  That seems like something that is more geared to a configuration module or some sort that is very specific to ones infrastructure.  I'd vote to close this issue and leave that responsibility up to a configuration wrapper module sort of thing.

aj-jester:
@split3 +1

solarkennedy:
Ok. @albustax the more standard way to approach this kind of need is with a profile, where you might have `profile_consul::check` that does a consul check as well as other site specific things, like deploying files, sudo rules, etc.

PR #128, opened 2015-05-28
Supply optional token for ACL changes

mdelagrange:
It seems likely that the Consul cluster will not allow the anonymous token to make ACL changes, so it would be nice if you could supply a token with the right privileges.  I tried setting that up in this PR.  I had to delete the prefetch methods, since they wouldn't have access to the token.  However maybe there is a better way to handle it.

### DIFF
--- a/lib/puppet/provider/consul_acl/default.rb
+++ b/lib/puppet/provider/consul_acl/default.rb
@@ -6,14 +6,7 @@ Puppet::Type.type(:consul_acl).provide(
 ) do
   mk_resource_methods
 
-  def self.instances
-    acls = list_resources
-    acls.collect do |acl|
-      new(acl)
-    end
-  end
-
-  def self.list_resources
+  def self.list_resources(acl_api_token)
     if @acls
       return @acls
     end
@@ -23,14 +16,14 @@ Puppet::Type.type(:consul_acl).provide(
     uri = URI('http://localhost:8500/v1/acl')
     http = Net::HTTP.new(uri.host, uri.port)
 
-    path=uri.request_uri + '/list'
+    path=uri.request_uri + "/list?token=#{acl_api_token}"
     req = Net::HTTP::Get.new(path)
     res = http.request(req)
 
     if res.code == '200'
       acls = JSON.parse(res.body)
     else
-      acls = []
+      raise(Puppet::Error,"Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
     end
 
     nacls = acls.collect do |acl|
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/consul_acl/default.rb
```diff
@@ -23,14 +16,14 @@ def self.list_resources
     uri = URI('http://localhost:8500/v1/acl')
     http = Net::HTTP.new(uri.host, uri.port)
 
-    path=uri.request_uri + '/list'
+    path=uri.request_uri + "/list?token=#{acl_api_token}"
```
solarkennedy: Does this work when the token is set to an empty string? Like consul tolerate a `/list?token=` ?

lib/puppet/provider/consul_acl/default.rb
```diff
@@ -53,18 +46,11 @@ def self.list_resources
     nacls
   end
 
-  def self.prefetch(resources)
```
solarkennedy: Hmm. I understand why you have to do this.
Let me think a bit about possible other solutions.

I'm a big fan of self.instances and prefetch. I think there might be a better way to gracefully degrade..

lib/puppet/provider/consul_acl/default.rb
```diff
@@ -23,14 +16,14 @@ def self.list_resources
...
-    path=uri.request_uri + '/list'
+    path=uri.request_uri + "/list?token=#{acl_api_token}"
     req = Net::HTTP::Get.new(path)
     res = http.request(req)
 
     if res.code == '200'
       acls = JSON.parse(res.body)
     else
-      acls = []
+      raise(Puppet::Error,"Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
```
solarkennedy: Is this going to be ok in bootstrapping mode where consul isn't running yet?

lib/puppet/provider/consul_acl/default.rb
mdelagrange: Yes, that's assumed to be the anonymous token.

lib/puppet/provider/consul_acl/default.rb
mdelagrange: It shouldn't be an issue, since it's no longer part of a prefetch in this version.

lib/puppet/provider/consul_acl/default.rb
mdelagrange: I've never written a custom provider, but it does seem better to keep the prefetch.  Maybe there's a way to configure a provider globally?

solarkennedy:
I am also not an expert on this part of the code (custom providers), but I understand that often the ACLs will not be setup in such a way for puppet to prefetch them.

I don't currently use this ACL code at all, but I would like to. Thank you for your help in getting this ACL code in a more production-ready state!

activars:
this commit seems to break "ensure" capability of creating ACL resources.

solarkennedy:
Crap :(

@activars can you pastebin with output?

PR #130, opened 2015-05-31
Switch to using start-stop-daemon for consul upstart init script

### LINKED ISSUE #126: GOMAXPROCS discarded by upstart init due to sudo's env_reset option
Hi,

In upstart init version GOMAXPROCS is set correctly but then the actual consul-agent process is started via sudo.
In Ubuntu the default sudo configuration has env_reset configured and GOMAXPROCS is not part of the env_keep configuration.
I'd suggest adding a default for the consul user which has GOMAXPROCS as part of env_keep configuration to avoid it being ignored.

bdellegrazie:
This avoids the use of sudo and permits passing the GOMAXPROCS setting
via the environment. In addition this change exposes the pid file again so
that other users can signal the agent.
Fixes #126

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -241,7 +241,8 @@ describe 'consul' do
     }}
     it { should contain_user('custom_consul_user').with(:ensure => :present) }
     it { should contain_group('custom_consul_group').with(:ensure => :present) }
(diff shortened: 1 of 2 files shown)


solarkennedy:
Thank you! As much as a I like the simplicity of the simple exec, I understand that the bugs are showstoppers.

PR #132, opened 2015-06-02
Escape the ID & make fixtures useable more widely

duritong:
* change fixture source to https to make it work behind corporate firewalls & proxies
* make it possible to use / within an id


solarkennedy:
Thanks!

PR #134, opened 2015-06-08
Add an exec to daemon-reload systemctl when the unit-file changes

robrankin:
When the unit-file changes we need to issue a 'systemctl daemon-reload'.   I've had to change the bin_dir parameter, which changes the unit-file, but currently doesn't notify systemd of this.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -109,9 +109,10 @@ class consul::config(
 
 
   exec { 'systemctl daemon-reload':
-    path      => ['/bin','/usr/bin','/sbin','/usr/sbin'],
-    command   => "systemctl daemon-reload",
-    subscribe => File['/lib/systemd/system/consul.service'],
+    path        => ['/bin','/usr/bin','/sbin','/usr/sbin'],
+    command     => "systemctl daemon-reload",
+    subscribe   => File['/lib/systemd/system/consul.service'],
+    refreshonly => true,
   }
 
 }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -110,7 +110,7 @@ class consul::config(
 
   exec { 'systemctl daemon-reload':
     path        => ['/bin','/usr/bin','/sbin','/usr/sbin'],
-    command     => "systemctl daemon-reload",
+    command     => 'systemctl daemon-reload',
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -40,6 +40,12 @@
           group   => 'root',
           content => template('consul/consul.systemd.erb'),
         }
+        exec { 'systemctl daemon-reload':
+          path        => ['/bin','/usr/bin','/sbin','/usr/sbin'],
+          command     => 'systemctl daemon-reload',
+          subscribe   => File['/lib/systemd/system/consul.service'],
```
solarkennedy: Instead of this line can you put a ~> on line 42 to chain them?

manifests/config.pp
```diff
@@ -40,6 +40,12 @@
           group   => 'root',
           content => template('consul/consul.systemd.erb'),
         }
+        exec { 'systemctl daemon-reload':
+          path        => ['/bin','/usr/bin','/sbin','/usr/sbin'],
```
solarkennedy: Is this in a predictable place? can we just use the fully qualified path?

manifests/config.pp
```diff
@@ -40,6 +40,12 @@
           group   => 'root',
           content => template('consul/consul.systemd.erb'),
         }
+        exec { 'systemctl daemon-reload':
```
solarkennedy: I'm sad we have to do this in general, but is is probable this will conflict with other modules.
https://github.com/camptocamp/puppet-systemd/blob/master/manifests/init.pp#L3
https://github.com/search?utf8=%E2%9C%93&q=%22exec+%7B+%27systemctl+daemon-reload%27%3A%22&type=Code&ref=searchresults

Can you take this approach and make the exec unique to this module?
https://github.com/garethr/garethr-docker/blob/master/manifests/systemd_reload.pp#L6

manifests/config.pp
robrankin: I don't know if its in the same place in every RH variant, or system that uses systemd.  Figured these were pretty likely paths.  Happy to change it.

robrankin:
Changes made, thanks for the suggestions.

solarkennedy:
Thanks!

PR #135, opened 2015-06-12
Make consul::install optional

potto007:
Added a check against $install_method for the string 'none'. If found, nothing is downloaded or installed.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -13,56 +13,57 @@ class consul::install {
     }
   }
 
-  if $consul::install_method == 'url' {
-
-    if $::operatingsystem != 'darwin' {
-      ensure_packages(['unzip'])
-    }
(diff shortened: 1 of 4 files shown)


manifests/install.pp
```diff
@@ -13,56 +13,57 @@
...
+      if $consul::ui_dir {
+        package { $consul::ui_package_name:
+          ensure => $consul::ui_package_ensure,
+        }
       }
     }
-
-  } else {
-    fail("The provided install method ${consul::install_method} is invalid")
+    'none': {}
```
solarkennedy: The readme is good for the full example, but can you update the inline docstring on init.pp:13?

potto007: Sorry about that, I overlooked the docstring. Updated.

solarkennedy:
Looks great! Thanks.

PR #137, opened 2015-06-16
Fix debian init

dizzythinks:
When running the Debian init_script manually the do_start function would immediately exit with a failure on the first test in the for loop sequence even if the process was actually running.

    for i in `seq 1 30`; do
        if ! start-stop-daemon --quiet --stop --test --pidfile $PIDFILE --exec $DAEMON --user $USER; then
            echo " FAIL: consul process died"
            return 2
        fi

Have altered this so it captures the return code and sleeps before proceeding through the for loop.

### DIFF
--- a/templates/consul.debian.erb
+++ b/templates/consul.debian.erb
@@ -52,7 +52,6 @@ do_start()
     #   0 if daemon has been started
     #   1 if daemon was already running
     #   2 if daemon could not be started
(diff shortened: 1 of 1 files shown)


solarkennedy:
Ug. I hate init scripts.

dizzythinks:
They can be a pain it's true! :-)

PR #138, opened 2015-06-17
Pass ensure to service definition file

mklette:
Adding `ensure` attribute, so already installed services can be removed again,


dominikschulz:
:+1:

PR #139, opened 2015-06-17
compatibiliy fix: ensure variables are defined

mklette:
Otherwise this code might break Puppet when configuration option
strict_variables is enabled.


dominikschulz:
:+1:

solarkennedy:
Failures on the future parser.
We also use strict variables in production, but I guess we don't exercise the code path that causes that failure.

Can you add STRICT_VARIABLES=true in the .travis.yml so we can be sure to catch failures like this in the future?

mklette:
Could you cross-check once again?

solarkennedy:
Great! Thank you.

PR #141, opened 2015-06-19
reorder resources for install_method => package

jlambert121:
This PR is in response to this problem:

Create consul package with fpm, set user/group to consul
Set data_dir to /var/consul/data

Problem:
Package cannot install because consul u:g are not created yet (can't be done with FPM packages ATM)
Puppet error creating /var/consul/data (parent dir doesn't exist)

Solution:
This really is only a partial solution, but I'm not sure how to make a better solution without some deep nasty conditionals with the different install methods.  This solution will work with the default "manifest" ordering (processing is top down), but will break if ordering is set to random.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -4,12 +4,14 @@
 #
 class consul::install {
 
-  if $consul::data_dir {
-    file { $consul::data_dir:
-      ensure => 'directory',
-      owner  => $consul::user,
-      group  => $consul::group,
-      mode   => '0755',
+  if $consul::manage_user {
+    user { $consul::user:
+      ensure => 'present',
+    }
+  }
+  if $consul::manage_group {
+    group { $consul::group:
+      ensure => 'present',
     }
   }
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
This "shouldn't" be a problem, as files auto-require the users and groups that own them:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file.rb#L345

I'm going to veto parse-dependent puppet code.

Is the problem more that we need to have puppet just ensuring the parent dir?

jlambert121:
It's both the parent dir (which in my case is created by the package) and the package tries to change the ownership of files to consul:consul (which don't exist when the package is installed).

The user:group issue I could see argued is a packaging problem (and I'd have to agree), but that doesn't solve the data dir issue.  

I agree parse-dependent solutions aren't ideal, but with the number of conditionals in the package it makes it hard to set dependencies correctly.  Would you be opposed to breaking the install options out into separate classes and then dependencies could be set at the class level?  (Option to other solutions too)

solarkennedy:
What if we simply manage the parent dir and the data dir?

jlambert121:
If puppet had a param to support 'mkdir -p' that would make a lot of these issues easier.

Do you mean just grab the parent dir of data_dir and ensure that exists as well?  That would solve my issue (other than the ordering problem for u:g still exists), but it would only work for people who have one level of parent dir and correct packages (which arguably I should write a spec file for to make correct ones, but FPM is so much easier).

solarkennedy:
Well, what if we solved this a different way and ensured the *user/group* existed before the package was installed? Would that also solve this problem?

jlambert121:
Is something like #145 what you were meaning?  All the conditionals feels "less puppet-like", but it is more explicit for sure (and not ordering dependent).

solarkennedy:
Yes. Closing in favor of #145

PR #142, opened 2015-06-19
create user/group as system accounts

jlambert121:
Service user/groups should be created as system accounts (IMO).  Not creating them as system accounts causes issues when using a directory service.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -69,11 +69,13 @@ class consul::install {
   if $consul::manage_user {
     user { $consul::user:
       ensure => 'present',
+      system => true,
     }
   }
   if $consul::manage_group {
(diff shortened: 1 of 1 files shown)


solarkennedy:
I agree. Can you make this not a boolean? I know this will probably break someone if they are running against master, but this is a mistake to not run these as system accounts.

jlambert121:
An even better solution IMO, just didn't want to take away an option someone wanted.  Updated

solarkennedy:
I ran a quick test on my own machine, puppet doesn't retroactively change the uid when that flag is set, it only uses it for creation time. This is fine.

solarkennedy:
Thanks!

PR #143, opened 2015-06-23
add puppet 4 testing to travis

jlambert121:


solarkennedy:
Thanks! I'm going to turn on strict vars everywhere and I think we should probably test against the future parser as well as not, in the later revisions of puppet.

PR #145, opened 2015-06-24
explicitly set depedencies for package install

jlambert121:


solarkennedy:
I think this is fine. Yes there are lots of ifs, but we pretty much have to because *everything* is optional here.

Now if you *really* want to be fancy, you could see your tests fail by writing an acceptance test to ensure this doesn't break in the future. We don't run the acceptance tests in travis as they are slow and require vagrant, but that would be the place to make sure this *actually* works.

PR #147, opened 2015-07-03
Adding groups parameter to user definition

robrankin:
Allows you to add the consul user to an array of groups, so the agent can for example execute watch handlers as consul but access/manage items on the filesystem that aren't owned by the consul user.


solarkennedy:
Seems legit.

PR #149, opened 2015-07-09
Modify consul_validate_checks to work with ruby 1.8

### LINKED ISSUE #148: Ruby 1.8 support
Ruby 1.8 isn't supported at the moment, as within https://github.com/solarkennedy/puppet-consul/blob/master/lib/puppet/parser/functions/consul_validate_checks.rb, it is trying to call 'Puppet::ParseError' as a function. Which results in 

```
undefined method `ParseError' for Puppet:Module at /etc/puppet/ext_modules/consul/manifests/service.pp:36 on node ip-172-18-0-234.ec2.internal
```

It appears that I can fix this by using `raise(Puppet::ParseError, 'message')` rather than `raise Puppet::ParseError('message')`, however I'm not a ruby developer so I'm not sure if this is the correct change to make or not.

Ruby 1.8 is the default ruby on Ubuntu 12.04 (which is why this is causing me issues). Would you be interested in a patch to fix support for 1.8 on this system configuration?

solarnz:
Fixes #148 

This adds ruby 1.8 support, which only involved changing the requirements in Gemfile and changing how exceptions are raised in consul_validate_checks.rb.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: ruby
 bundler_args: --without development
 before_install: rm Gemfile.lock || true
 rvm:
+  - 1.8.7
   - 1.9.3
   - 2.0.0
   - 2.1.0
--- a/Gemfile
+++ b/Gemfile
@@ -3,13 +3,18 @@ source "http://rubygems.org"
 group :test do
   gem "rake"
   gem 'beaker', '~> 1.11.0'
-  gem "puppet-blacksmith"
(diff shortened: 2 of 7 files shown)


solarkennedy:
Great!

solarkennedy:
Sorry, I had to undo the travis part of this, as we were not properly installing the correct puppet version before.

The rest of your changes are still in place though.

PR #153, opened 2015-07-15
travis update

jlambert121:
In an effort to speed up travis testing:
Removed some intermediate puppet versions that weren't in PE
Enabled travis's container-based testing
FUTURE_PARSER also includes STRICT_VARIABLES (and neither parameter exists in 4.x)

In all the travis tests I've watched run errors are typically either on all tests, old version of ruby, or future parser/strict_variables.  Testing every version seemed like it just made things take longer without any real benefit, your experience may be different though.  (I'd also vote to drop 2.7 testing/support - it's 2 major versions behind)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: ruby
 bundler_args: --without development
 before_install: rm Gemfile.lock || true
+sudo: false
 rvm:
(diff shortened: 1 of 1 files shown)


solarkennedy:
Thanks!

PR #155, opened 2015-07-16
Correct sysv init script to check if sockets are free. This solve pro…

bartlomiejsawicki:
…blem during restart when old process stop but sockets are still in use and then new one can't start.
I found that consul agent after stop sometimes is leaving open connection which can last for me up to 45 seconds (this is the highest value which I found during tests). This fix is solving the problem by adding the loop which check if sockets are free and after that is finishing stop function in init script. So when we try to start again we are sure that process will not crash.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -13,8 +13,28 @@
 class consul::config(
   $config_hash,
   $purge = true,
+  $port_hash = undef,
 ) {
-
+  $port_hash_default = {
+    'dns'      => 8600,
+    'http'     => 8500,
+    'https'    => -1,
+    'rpc'      => 8400,
+    'serf_lan' => 8301,
+    'serf_wan' => 8302,
+    'server'   => 8300,
+  }
+  $config_ports_not_int = merge($port_hash_default, $port_hash)
+  $config_ports={
+    'dns'      => $config_ports_not_int['dns'] * 1,
+    'http'     => $config_ports_not_int['http'] * 1,
+    'https'    => $config_ports_not_int['https'] * 1,
+    'rpc'      => $config_ports_not_int['rpc'] * 1,
+    'serf_lan' => $config_ports_not_int['serf_lan'] * 1,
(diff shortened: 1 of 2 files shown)


solarkennedy:
This is a pretty complicated way to solve this.

Can you solve this with less complexity by simply waiting for the pid to die?

bartlomiejsawicki:
The problem is that pid is dead but socket is still opened without process
assign to it. So init script thinks that everything is ok.
16 lip 2015 16:42 "Kyle Anderson" <notifications@github.com> napisał(a):

aj-jester:
@bartlomiejsawicki Hey, trying to understand this problem a little better, can you provide some more information?

* In `sysctl.conf` can you put `net.ipv4.tcp_tw_reuse = 1` (don't forget to run `sysctl -p`) and then try restarting via the init script a few times?
* What is your TCP `TIME-WAIT` timeout?
* You mentioned this happens sometimes, out of 10 restarts how many take more than a few seconds?
* Does it happen during certain config changes? main config (config.json) vs service/check/watch changes? or just randomly?
* Can you try replicating this issue with the agent directly, something like this (ensure you pass the same options as the init script):
```
# /usr/local/bin/consul agent -pid-file /var/run/consul/consul.pid -config-dir <condif_dir> <extra options>
^C
# /usr/local/bin/consul agent -pid-file /var/run/consul/consul.pid -config-dir <condif_dir> <extra options>
^C
# /usr/local/bin/consul agent -pid-file /var/run/consul/consul.pid -config-dir <condif_dir> <extra options>
^C
...
```

Some jargon on all this:
http://vincent.bernat.im/en/blog/2014-tcp-time-wait-state-linux.html
http://www.lognormal.com/blog/2012/09/27/linux-tcpip-tuning/
https://rtcamp.com/tutorials/linux/sysctl-conf/

bartlomiejsawicki:
I think we can close it, sorry for answering so late but new version of your branch was solving the problem so this change is not needed.

PR #157, opened 2015-07-17
remove string casting to int

aj-jester:
@solarkennedy now that we have merged https://github.com/solarkennedy/puppet-consul/pull/156 we can remove all this string casting business :smile:


solarkennedy:
Ah. So great :)

PR #158, opened 2015-07-23
UI dir symlink should depend on the dist existing

jsok:
I've had some issues with the ui dir symlink when trying to set the data dir to a path on an LVM volume that doesn't (yet) exist.

Even with `Lvm::Volume_group <||> -> Class['consul']` the symlink creation fails. Which has me a little baffled.

Either way, this fix is trivial and probably makes sense to delay the symlink creation anyhow!

Trace:
```
(/Stage[main]/Staging/File[/opt/staging]/ensure) created
(/Stage[main]/Consul::Install/Staging::File[consul.zip]/File[/opt/staging/consul]/ensure) created
(/Stage[main]/Consul::Install/Staging::File[consul.zip]/Exec[/opt/staging/consul/consul.zip]/returns) executed successfully
(/Stage[main]/Consul::Install/Group[consul]/ensure) created
Could not set 'link' on ensure: No such file or directory - /var/consul at 49:/etc/puppet/modules/consul/manifests/install.pp
Could not set 'link' on ensure: No such file or directory - /var/consul at 49:/etc/puppet/modules/consul/manifests/install.pp
Wrapped exception:
No such file or directory - /var/consul
(/Stage[main]/Consul::Install/File[/var/consul/ui]/ensure) change from absent to link failed: Could not set 'link' on ensure: No such file or directory - /var/consul at 49:/etc/puppet/modules/consul/manifests/install.pp
```

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -42,7 +42,7 @@ class consul::install {
           source  => $consul::real_ui_download_url,
           target  => "${consul::data_dir}/${consul::version}_web_ui",
           creates => "${consul::data_dir}/${consul::version}_web_ui/dist",
-        }
+        } ->
         file { $consul::ui_dir:
           ensure => 'symlink',
           target => "${consul::data_dir}/${consul::version}_web_ui/dist",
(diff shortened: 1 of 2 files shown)


aj-jester:
@jsok Can we please add a test for this?

jsok:
I've added an assertion that the relationship exists and that it's a symlink

solarkennedy:
Thanks!

aj-jester:
@jsok Looks good! :smile:

PR #159, opened 2015-07-31
Pretty json config

teaforthecat:
render pretty json into /etc/consul/config.json

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -169,6 +169,7 @@ GEM
       signet (~> 0.6)
     hiera (1.3.4)
       json_pure
+    hiera-puppet-helper (1.0.1)
     hocon (0.9.0)
     http-cookie (1.0.2)
       domain_name (~> 0.5)
@@ -279,6 +280,9 @@ PLATFORMS
 DEPENDENCIES
   beaker (> 2.0.0)
   beaker-rspec (>= 5.1.0)
+  hiera
+  hiera-puppet-helper
+  json
   pry
   puppet (~> 3.7.0)
   puppet-blacksmith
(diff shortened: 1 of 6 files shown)


aj-jester:
Hmmm, whats the reason for this?

teaforthecat:
I thought it would be nice to read /etc/consul/config.json with newlines in it. It's more readable.

teaforthecat:
I'll add some newlines to the other tests that I missed

aj-jester:
I'm assuming the only time someone would need to read the config is while troubleshooting.

Why not do something simple one-liner like:
```
ruby -e "require 'json'; puts (JSON.pretty_generate JSON.parse(IO.read('/etc/consul/config.json')))"
```

This way we keep the codebase lean. @solarkennedy thoughts?

teaforthecat:
We're dealing with system administrators here. Asking them to run that command above to look at a file would produce some funny looks. I understand the ideal situation is where no one logs onto the server at all. In reality, though it happens a lot, and it would be nice to have a readable configuration file when someone does want to take a look. Also, imho, this version is a little easier to read.

solarkennedy:
`cat file.json | jq .`

I was sad when I had to write this function in the first place, I don't really want to see it grow.

Can't we just stick our own JSON.pertty_generate() on our object in the function at the last second when we return?

aj-jester:
One of my biggest concerns with this is whitespace related changes. I don't think anyone wants a restart because of it :worried:.

If we do decide to implement this, can we at least make it conditional with "pretty" being turned off by default?

split3:
I agree, this should at least be conditional.

solarkennedy:
With two votes for it being a conditional option, I'm going to veto adding this to the module. For pretty printing any json file I recommend `jq . your.json`

solarkennedy:
Sorry :(

PR #160, opened 2015-07-31
[install] use data_dir from config_hash

teaforthecat:
the data_dir variable in install.pp appears to be unused making it
impossible to change the owner of the data_dir directory. This makes it possible to change the owner of the data_dir, as well as force consistency with the configuration.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -123,7 +123,10 @@ class consul (
 
   anchor {'consul_first': }
   ->
-  class { 'consul::install': } ->
+  class { 'consul::install':
+    data_dir => $config_hash_real['data_dir'],
+  }
+  ->
   class { 'consul::config':
     config_hash => $config_hash_real,
     purge       => $purge_config_dir,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -2,10 +2,12 @@
 #
 # Installs consule based in the parameters from init
 #
-class consul::install {
+class consul::install (
+  $data_dir = undef,
+  ){
 
-  if $consul::data_dir {
-    file { $consul::data_dir:
+  if $data_dir {
+    file { $data_dir:
       ensure => 'directory',
       owner  => $consul::user,
       group  => $consul::group,


aj-jester:
Seems to be working fine when I test is against the latest master :confused: 

Puppet run with group changed:
```
Notice: /Stage[main]/Consul::Install/File[/opt/consul]/owner: owner changed 'consul' to 'root'
Notice: /Stage[main]/Consul::Install/File[/opt/consul]/group: group changed 'consul' to 'root'
```

```
Notice: /Stage[main]/Consul::Install/File[/opt/consul]/owner: owner changed 'root' to 'consul'
Notice: /Stage[main]/Consul::Install/File[/opt/consul]/group: group changed 'root' to 'consul'
```

solarkennedy:
Yea, I don't mind moving the data_dir to be injected into the install class, but if you do that, please remove that other plumbing that aj linked to.

I does look like we don't have tests coverage for ensuring that the owner is set correctly. So if isn't, please add a test to ensure it is correct and stays correct.

aj-jester:
@solarkennedy I wouldn't do that because this [block](https://github.com/solarkennedy/puppet-consul/blob/master/manifests/init.pp#L104-106) requires it. Although a test to ensure proper user/group seems like a fantastic idea!

I guess Im trying to understand why we think the existing code isin't working? Is there a way we can replicate it?

teaforthecat:
I wonder if the version of puppet makes a difference here. I'm running `3.2.4 (Puppet Enterprise 3.0.1)`. Are you on a different version?
I don't see how `$consul::user` in the `install.pp` resolves to the value in the `init.ppl`, so maybe that is new.

aj-jester:
I tested this on `3.3.2`. The way puppet inheritance works, because we call `class { 'consul::install': }` in `init.pp`, `install.pp` has access to all params from `init.pp`.

split3:
Is it safe to close this pull request now?  Seems that there isn't actually a problem here instead of there being no tests explicitly against the user/group ownership.

teaforthecat:
Yes, sorry for the lack of response. We may revisit this at a later time, but we aren't using this module at this time. It is probably something on our end. Thank you.

PR #163, opened 2015-08-08
Adding support for Ubuntu 15.04

asasfu:
Added fact 'is_systemd' to detect if Ubuntu 15.04 is running in upstart or systemd mode, systemd is default but still optional.
Added support in params to detect Ubuntu 15.04.
Added support for Ubuntu 15.04 path to systemctl by changing how the path ENV is handled in config.pp (Ubuntu 15.04 uses /bin/systemctl)


manifests/params.pp
```diff
@@ -29,9 +29,21 @@
...
       $init_style = 'debian'
-    } else {
+    } elsif versioncmp($::operatingsystemrelease, '15.04') < 0 {
       $init_style = 'upstart'
     }
+    else {
+      # Introducing fact 'is_systemd' since Ubuntu 15.04 can be systemd OR upstart (defaults to systemd though)
+      # Fact allows us to retrieve the client side value.
+      # We convert to boolean incase some people may have stringify facts turned on or not set in puppet 3.x still.
+      $is_systemd = str2bool($::is_systemd)
```
aj-jester: can we condense this into:

```
$init_style = str2bool($::is_systemd) ? {
  true    => 'systemd',
  default => 'upstart',
}
```

asasfu:
Sure can.  I'll do that in a few minutes.  I'm working through your code at the moment as well to patch it up as It doesn't update consul version if it's already installed.  This is due to the 'creates' section in staging and how this module is using it.  I have a plan of attack to make use of versioning and am about half way through patching it... well fully done with consul, investigating web_ui at the moment.

manifests/params.pp
asasfu: Sure can. I'll do that in a few minutes. I'm working through your code at the moment as well to patch it up as It doesn't update consul version if it's already installed. This is due to the 'creates' section in staging and how this module is using it. I have a plan of attack to make use of versioning and am about half way through patching it... well fully done with consul, investigating web_ui at the moment.

manifests/params.pp
```diff
@@ -29,9 +29,21 @@
   if $::operatingsystem == 'Ubuntu' {
     if versioncmp($::lsbdistrelease, '8.04') < 1 {
       $init_style = 'debian'
-    } else {
+    } elsif versioncmp($::operatingsystemrelease, '15.04') < 0 {
```
aj-jester: Can we either use `$::lsbdistrelease` here or replace the above with `$::operatingsystemrelease` for consistency sake? I'm in favor of the latter because `$::operatingsystemrelease` is used everywhere else.

asasfu: I can't be authoritative on that comment.  I don't know Ubuntu that far back to know if puppet supported $::operatingsystemrelease fact back then, I've heard that there was no $::operatingsystemmajrelease in older versions of some linux distros.  So I figured i'd let it stay, just in case as i'm not the pro on old linux systems.

manifests/params.pp
aj-jester: Oh nice! This has been a LONG standing issue with the url install method. I myself have been wanting to write one just hadn't had the time.

This is as close as we've gotten https://github.com/solarkennedy/puppet-consul/pull/118

asasfu: Ya, I saw that.  I'm pretty sure mine will work, theoretically it does :)  But I may call it a weekend before I get any thorough testing done on it and return to fixing it on Monday.  I'm hoping to deploy consul using this to our University soon enough so I want to make sure everything is up to snuff.  Since there is no yum packages or apt that I know of for consul in the distros that we're using.

manifests/params.pp
aj-jester: I wasn't aware. Then lets change this to `$::lsbdistrelease` as well since we know it exists even on the latest versions. My concern is we don't want to troubleshoot a bug where these two facts have different values.

asasfu: I started programming puppet after version 3 and I've been a fan of less dependencies.  So I was always trying to convert our old stuff to operatingsystemrelease and operatingsystemmajrelease.  
lsbdistrelease is relying on redhat-lsb and whatever the package is in Ubuntu and other distros.  I believe the operatingsystem values are coming straight from facter Facter::Operatingsystem.implementation
Though i'm not sure where it gets that from.  I'm pretty sure the standing thought in puppet programming was to phase lsb out but that might be a personal preference of people...  as I said, I haven't been doing this long enough. :)

aj-jester: According to https://launchpad.net/ubuntu/vivid/+package/lsb-release

> "The Linux Standard Base (http://www.linuxbase.org/) is a standard
 core system that **third-party applications written for Linux can
 depend upon**"

So I think we are safe :smile: my vote is for consistency over deprecation/dependencies, in this specific case. But I agree with you in general, we should try to move away from legacy stuff as much as possible.

aj-jester: Also, if you are curious, this is how `$::operatingsystemrelease` gets its fact: https://docs.puppetlabs.com/facter/1.6/core_facts.html#operatingsystemrelease
https://docs.puppetlabs.com/facter/2.3/core_facts.html#operatingsystemrelease

As you can see its a mess lol

aj-jester:
@asasfu I'm not an expert in `systemd` but can you double check to see if we need to add a new [$init_selector](https://github.com/solarkennedy/puppet-consul/blob/master/manifests/run_service.pp#L8-10) for `systemd` called `consul.service` like so:

```
$init_selector = $consul::init_style ? {
    'launchd' => 'io.consul.daemon',
    'systemd' => 'consul.service',
    default   => 'consul',
  }
```

Please check this PR for more context https://github.com/solarkennedy/puppet-consul/pull/125

manifests/params.pp
```diff
@@ -29,9 +29,18 @@
...
       $init_style = 'upstart'
     }
+    else {
```
aj-jester: @asasfu What do you think about actually getting rid of the fact `is_systemd` and making 15.04 or higher default `systemd` like the rest of the operating systems? The reason being `params.pp` lists the "default" for an OS, in this case the default for 15.04 or higher being `systemd`. If a user chooses to use `upstart` then they can override it via `$consul::init_style`.

I would like to hear @solarkennedy thoughts on this as well.

```
if versioncmp($::lsbdistrelease, '8.04') < 1 {
  $init_style = 'debian'
} elsif versioncmp($::lsbdistrelease, '15.04') < 0 {
  $init_style = 'upstart'
} else {
  $init_style = 'systemd'
}
```

asasfu: It's a thought.  I know at our University we have a computing science department with a good amount of 15.04s running upstart, but with plans to move them to systemd once they get a better understanding of it and integrate it.  We're still on puppet 3 with plans to jump into puppet 4 soon enough.  Puppet 3.x doesn't support fedora 22 or Ubuntu 15.04 so I wrote the is_systemd fact specifically for Ubuntu 15.04 so I could put that into my sites.pp file to avoid it thinking that it was upstart still.  Fedora 22 didn't need the fact but still needed to get a Service { provider => 'systemd' } due to it not being in puppet 3.x

The downside to doing overrides on the init_style is that you can't have a mixture of upstart and systemd Ubuntu 15.04 in the same role/profile or hostgroup(foreman)  Otherwise you'd have to put in a few more snowflakes.

aj-jester: To get a mixture while you guys are transitioning why not to do something like below?

```
$init_type = $::is_systemd ? {
  true    => 'systemd',
  default => 'upstart',
}

class { '::consul':
  init_type => $init_type,
}
```

I guess I don't understand why `is_systemd` should be a fact when we don't have `is_upstart`, `is_launchd`, `is_sysv`, etc. facts. So that makes me think we are headed the wrong direction with this :confused:

asasfu: Mainly because systemd is now being transitioned to and Ubuntu was annoying about it.  I figured it might be helpful to everyone.  It may or may not be good to have in a module rather than in a private wrapper module which we would load on the side to do that suggested change.  At least the fact may likely help someone :) whether in this module or not.   And yes we have a facts module for our own Uni. that contains it as well.

manifests/params.pp
aj-jester: No rush dude! heh. Yah, unfortunately there are no packages. When I have some time I'll build a package and try to deploy it to a public apt repo.

Let me know how the consul deploy goes!

asasfu: Hey AJ.  
You've been really helpful so far with the responses to my PR.  You mind just peeking at this commit with the working consul upgrade/downgrade (web_ui) not done yet but part way in.  
https://github.com/sfu-rcg/puppet-consul/commit/8fbbf44266768e311cc6cf0ca4874c8aa39d45e4
It tests fine for me.  The Exec { provider => shell } is because staging module doesn't allow us to push path to the exec and I can't tie the test, grep, and rm down to specific directories in the 'onlyif' statement as they are located in different locations in Ubuntu vs CentOS.  provider => shell fixes that up :)

manifests/params.pp
solarkennedy: I would prefer a "tell don't ask" approach here. What if we default to the stock init system of the operating systems we support, and if users want to force a different one, they can via the class param.

That is what the params class is for: providing sane defaults on a per-environment basis, but allowing users to override them. Because in theory the same thing could be happening on debian, or arch, etc.

asasfu: That's true, you're probably both right.  I've often thought of puppet as trying to cover all the bases by default to reduce the amount of conditionals that have to exist outside of the puppet module itself when it comes to OS compatibility, but at the same time, this may be one abstraction too far as it's likely a fairly edge case where I was trying to support the outlying resistant to systemd crowd while it is only a few and this module already exposes the option to override $init_style, so it isn't surgery to make it work from the outside for the stragglers.  I'll clear up the is_systemd stuff.  I'm also going to bump the lsbdistrelease over to operatingsystemrelease for Ubuntu 8's line as that value has been around long enough in facter, it was just operatingsystemmajrelease that wasn't it seems.

aj-jester:
@asasfu this looks great! Can you please write a unit test for this? After that this should be good to merge!

asasfu:
I'm not sure what happened on that Travis CI build but it kinda went tits up...   I have never done a unit test before... :(

asasfu:
Also, i'm curious, how do you guys handle the 'consul reload' .... if the ENV variable(s) are not set such as CONSUL_RPC_ADDR=`facter ipaddress`:8400  then `consul reload` fails during the puppet run...

aj-jester:
All params are set in `config.json`, nothing is via ENV.

In this case, if you want a specific rpc address just pass the `rpc` key in `$consul::config_hash` (this will basically convert the hash into a json object and dumps it in `config.json`. So you can pass any configuration option supported by consul).

If you dont set `rpc` key or if ENV var `CONSUL_RPC_ADDR` is not set, it defaults to `client_addr` https://www.consul.io/docs/agent/options.html#rpc

asasfu:
Ah you're right, it doesn't affect the puppet run's consul reload, just the user if he wants to type `consul members` or something needs to set his ENV or set client_addr to 0.0.0.0 or use reverse_proxy... because if client_addr is set to the specific IP it seems that when you run `consul members` from the command line it tries to contact 127.0.0.1:8400

asasfu:
Ok, my commit history time issue was corrected and I fixed up the spec tests.  This should be ready for production.  

On a side note: When Puppet loads all of its providers and checks its confines to decide which Service provider matches your system.  Do you guys know if there's a way to `get` the provider that it picked via the confines?  That'd be really nice.  Since their confines do all the work and then it's just one place to track that information down from in order to decide what init_style puppet has decided to be correct for your system, based on confines that they maintain.

aj-jester:
:bullettrain_side: LGTM

> Do you guys know if there's a way to `get` the provider that it picked via the confines?

I'm not exactly sure if thats possible :confused:

solarkennedy:
Thanks!

Yea you can't get a provider for something in puppet and have that data influence the catalog; the catalog is compile first and *then* applied to the host where providers are determined at the last moment.

asasfu:
Hey hey, you wanna know something you can do?!  While i'm preparing the ability for URL upgrade/downgrade I finally figure out how to stop the service before removing consul bin.  

service_name = 'consul'
actionstop = 'start' # e.g., (start|stop|restart|status)
actionstart = 'start' # e.g., (start|stop|restart|status)

require 'puppet'
service = Puppet::Type.type(:service).newservice(:name => service_name)

service.provider.send(actionstop)

# REMOVE CONSUL BIN

service.provider.send(actionstart)

So I just need to create a function that does all the work and does those steps as well.

asasfu:
@aj-jester and @solarkennedy you mind looking at this commit to see if you have any suggestions or gripes.  I know it has some conditionals in it but it works for all cases that should be presented.  It also leaves staging files behind per version rather than cleaning up, something I can add(cleanup) if wanted.
https://github.com/sfu-rcg/puppet-consul/commit/707af3f303be931bfda112fc1c9117b58d0ec60c

asasfu:
My work is likely considered complete on that patch.  Can you guys take a peek.  I'll be submitting it as a pull request tomorrow.  It has tested good on both my masterless and mastered puppet client.  I'm not sure what or any tests you may want written for this prior to merging?

asasfu:
Also now that I've done my testing, I've finally come across what I was looking for, that might be helpful to you guys when doing up params to decide init_style.
  $prov = inline_template("<%= Puppet::Type.type(:service).newservice(:name => 'nonexistant').provider %>")
  $provsub = split($prov, 'provider=([A-Za-z]+)')
  notify { "Provider is: ${provsub[1]}": }

again though, this only supports providers already setup in confines by puppet, this doesn't support any site.pp overrides for provider or anything, so those slight edge cases like Ubuntu 15.04 and Fedora 22 in Puppet 3.x will still return the wrong provider as this is ASKING puppet what they think the provider should be.  But still altogether, really cool.

asasfu:
Ok part 2, even better.  Provider with and without override provided by an upper class or site.pp via:
Service { provider => 'systemd' }

With override, what I'm doing here is calling to defaults and return provider:
  $defaultprovscoperbe = inline_template("<% @def = scope.lookupdefaults('Service') %><%= @def[:provider] %>")

Without override, here I'm retrieving service provider with no allowance for anyone to have overridden it earlier and they'd have to use consul::init_style to actually override it then for example:
  $prove = inline_template("<% @def2 = Puppet::Type.type(:service).newservice(:name => 'anyservice') %><%= @def2[:provider] %>")

Finally, no parsing needed this time, tested and showing proof:
  notify { "Default Provider is: ${defaultprovscoperbe} and Provider is: ${prove}": }

asasfu:
Last part, part 3.

This is the final piece of code as I noticed a parse needed to happen on the default provider on part2 and then this supports merging.

If you provide a Service { provider => systemd } in your site.pp or even parent class calling consul then you would override service { 'consul': } unless it was service { 'consul': provider => 'upstart' } as the more specific class wins if it defines that parameter or hash key/value.  
I've provided the way to do this via ruby so that you can decide which one you would like, the one that is overridden by the user via Service { provider => systemd } or the one that doesn't care if you overrode it at all, it takes only what puppet's confines say the provider is.
I highly prefer the overridden one for services because you should probably always allow the user to have the final say on the provider if they needed to specify it in the first place then they've decided puppet isn't handling their provider via confines and needs fixing. 


  $merged_override_wins = inline_template("<% @puppet_decided = Puppet::Type.type(:service).newservice(:name => 'anyservice')[:provider].to_s %><% @override = scope.lookupdefaults('Service')[:provider].instance_variable_get(:@value) %><% @override = @puppet_decided if @override.nil? || @override.empty? %><%= @override %>")

  notify { "Merged provider with override being the winner: ${merged_override_wins}": }

$non_overridden = inline_template("<%= Puppet::Type::Service.new(:name => 'anyservice')[:provider] %>")

  notify { "Provider is: ${non_overridden}": }

Might be a consideration to use this code to decide your init_style rather than what is currently in params.  Then you would be modifying your conditional to match what they have in /usr/lib/ruby/vendor_ruby/puppet/provider/service/
 /usr/lib/ruby/vendor_ruby/puppet/provider/service/
base.rb         daemontools.rb  freebsd.rb      init.rb         openbsd.rb      openwrt.rb      runit.rb        smf.rb          systemd.rb      windows.rb
bsd.rb          debian.rb       gentoo.rb       launchd.rb      openrc.rb       redhat.rb       service.rb      src.rb          upstart.rb

Then you'd just have a selector stating what init script you're building based off that.  No need to say, centos < 7 is redhat and centos >= 7 is systemd, ubuntu < 8 is debian, ubuntu > 8 is upstart, ubuntu > 15 is systemd.  Kinda seems like a lot of decisions we're making that puppet or the user already does for us now that we know how to grab their decisions.

Thoughts?

PR #165, opened 2015-08-10
[consul_sorted_json] render pretty json

teaforthecat:
but only if $config::pretty_print_json is true for backwards compatibility


teaforthecat:
this is a rehash of #159 but with the default value of "pretty" false for backwards compatibility

lib/puppet/parser/functions/consul_sorted_json.rb
```diff
@@ -1,43 +1,91 @@
...
+
+  module_function
+
+
+  def sorted_json(config_hash, pretty)
+    cleaned = convert_integers( config_hash )
+    sorted  = sort_keys( cleaned )
+
+    if pretty
+      JSON.pretty_generate( sorted )
```
aj-jester: You are doing post-processing (making is pretty) after sorting (line `10`). How are you ensuring `pretty_generate` and `generate` will still keep the keys in order? As you might already know, ruby does *not* guarantee keys order once you serialize a hash to JSON.

http://stackoverflow.com/questions/5517007/json-parse-changes-sorting-ruby/5518671#5518671

teaforthecat: yeah, that's been a concern. That concern isn't related to this pull request though - at least I don't how that concern would change from this implementation to the current one (to_json).

aj-jester: Its not the same. If you notice, in the currently implementation we only do `to_json` on **non-array** and **non-hash** objects, for arrays and hashes we do string concatenation to ensure order, and once we sort we do *zero* post-processing. But here you are processing the entire hash, which will *not* guarantee order :smile:

teaforthecat: oh good catch, sorry I didn't see that. I'll close this ticket and rethink the approach.

teaforthecat:
This implementation doesn't ensure the order of the keys.

aj-jester:
I have a suggestion if you want to rework this. Can we create 2 separate methods in `consul_sorted_json.rb`; `def sorted_json` and `def pretty_sorted_json`. Keep the current code in `def sorted_json` and pretty in `def pretty_sorted_json`. There might be some repeat but this will make it easy for troubleshoot in the future; we'll know exactly what function is being used when `pretty` is true or false.

Thanks!

manifests/init.pp
```diff
@@ -35,12 +35,17 @@
...
+# [*pretty_print_json*]
+#   Add whitespace to config to make it more readable. Defaults to `false`.
+#
 class consul (
   $manage_user           = true,
   $user                  = 'consul',
   $manage_group          = true,
   $extra_groups          = [],
   $purge_config_dir      = true,
+  $print_pretty_json     = false,
```
aj-jester: For consistency sake, can we name all these vars `pretty_config` (in `config.pp` as well)? Currently I see `pretty_print_json`, `print_pretty_json` and `pretty` lol.

manifests/config.pp
```diff
@@ -91,7 +95,7 @@
   file { 'consul config.json':
     ensure  => present,
     path    => "${consul::config_dir}/config.json",
-    content => consul_sorted_json($config_hash),
+    content => consul_sorted_json($config_hash, $pretty),
```
aj-jester: We should do this change for `service`, `check` and `watch` as well for consistency.

spec/functions/consul_sorted_json_spec.rb
```diff
@@ -1,13 +1,51 @@
...
-  it { should run.with_params({
-    'w' => -1,
-    'x' => '8500',
-    'y' => '-8656',
-    'z' => 'foo bar 123 4 5 6',
-    }).and_return('{"w":-1,"x":8500,"y":-8656,"z":"foo bar 123 4 5 6"}')
-  }
+describe 'consul_sorted_json', :type => :puppet_function do
+
+  let(:test_hash){ { 'z' => 3, 'a' => '1', 'p' => '2' } }
```
aj-jester: please check for negative ints as well, quoted `'-1'` and unquoted `-1`.

manifests/config.pp
```diff
@@ -10,9 +10,13 @@
 # [*purge*]
 #   Bool. If set will make puppet remove stale config files.
 #
+# [*pretty_print_json*]
+#   Add whitespace to config to make it more readable. Defaults to `false`.
+#
 class consul::config(
   $config_hash,
   $purge = true,
+  $pretty = false,
```
aj-jester: I think it would be better if this is not parameterized but rather inherited from `init.pp` so that we can use the same value across all configs (services, checks and watches). We should avoid some configs being pretty while other not.

In other words, something like this:
```
consul_sorted_json($config_hash, $consul::pretty_config)
```

manifests/init.pp
teaforthecat: All good ideas. Thanks for the feedback.

aj-jester:
@teaforthecat https://github.com/solarkennedy/puppet-consul/pull/175 :grin:

PR #167, opened 2015-08-13
Fix url upgrade and downgrade

asasfu:
This would allow support of upgrading and downgrading via URL method.  Version would be a required parameter only to upgrade or downgrade, if not supplied the file will just be for example, '_redhat_amd64.zip', not an issue.

I have had talks with puppetlabs devs and puppetlabs-transition module is meant to help with this but it is still version 0.1.0 and I haven't seen it be maintained since earlier this year; so in the mean time I decided to go with the ruby calls.  I haven't tested yet if the transition module supports overrides like:
Service { provider => 'systemd' } or not.  In this current iteration I've not put support in here for it either, only conditional logic to support init_style changing the provider on the ruby call in the case of two unsupported distros in puppet 3.x
I accept discussions on this of course.  We can also instead of a ruby call, use the slightly slower call of:
puppet resource service consul provider=systemd ensure=stopped

Cheers,

Adam

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -27,8 +27,9 @@ class consul::install {
       ##  after it gets some patchwork and closer to release version
       #
       # Puppetversion fact has to be in quotes due to versioncmp differences between Puppet 3.x and 4.x
+      $puppet_version = inline_template("<%= Facter.value('puppetversion').to_s %>")
       # This was done for puppet 3.x not supporting Ubuntu 15 and Fedora 22, and since this ruby line doesn't support site.pp... override of Service { provider => 'systemd' }
-      if versioncmp("${::puppetversion}", 4.2) < 0 {
+      if versioncmp($puppet_version, 4.2) < 0 {
         $ruby_service_stop = $::operatingsystem ? {
           'fedora' => "ruby -r 'puppet' -e \"Puppet::Type.type(:service).newservice(:name => 'consul', :provider => '${consul::init_style}').provider.send('stop')\"",
           'ubuntu' => "ruby -r 'puppet' -e \"Puppet::Type.type(:service).newservice(:name => 'consul', :provider => '${consul::init_style}').provider.send('stop')\"",
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -26,13 +26,13 @@ class consul::install {
       } ->
       staging::extract { "${consul::real_download_file}":
         target         => $consul::bin_dir,
-        onlyif         => "test `consul version | grep -m1 -o [0-9\\.] | tr -d '\\n'` != ${consul::version}; unlessval=$?; if [ \$unlessval = 0 ]; then rm -f ${consul::bin_dir}/consul; fi; test \$unlessval = 0",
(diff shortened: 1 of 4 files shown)


asasfu:
Love it or leave it, I just gotta say puppet's versioncmp is a pain in the ass when they switched over to 4.x.  Plus I needed that ENV['PUPPET_VERSION'] in the init_spec for my fact puppetversion.
Either way, feel free to discuss this stuff with me.  I know it looks ugly but it really does work.  Could probably use some rspec tests, not 100% sure what to write for those specific cases, something like var should equal 'certain text'...

manifests/install.pp
```diff
@@ -15,21 +15,47 @@
...
-      staging::file { 'consul.zip':
+      staging::file { $consul::real_download_file:
         source => $consul::real_download_url
       } ->
-      staging::extract { 'consul.zip':
-        target  => $consul::bin_dir,
-        creates => "${consul::bin_dir}/consul",
+      staging::extract { $consul::real_download_file:
+        target => $consul::bin_dir,
+        unless => "which consul > /dev/null ; if [ $? = 0 ]; then test `consul version | grep -m1 -o [0-9\\.] | tr -d '\\n'` = ${consul::version}; unlessval=$?; if [ \$unlessval = 1 ]; then ${clean_staging_dir}; ${ruby_service_stop}; rm -f ${consul::bin_dir}/consul; fi; else unlessval=1; fi; test \$unlessval = 0",
```
solarkennedy: While I appreciate the effort to solve this problem, this is not the way to do it.

We are mostly working around the limitations of the staging module, what if we used this module instead:
https://github.com/camptocamp/puppet-archive

aj-jester: I agree with @solarkennedy on this. I think we should go with `puppet-archive` which will also allow us to do other cool things like `checksum` among other things.

asasfu: In some ways, I agree, I'd love to have a more pretty way.  Have you guys noticed...  Consul servers are very touchy about their quorum and regaining it tends to still require manual intervention.  I'm not even sure if puppet in this current consul release can manage the consul servers...only clients
I still have to review puppet-archive as the checksum method is nice.  Still should have a service stop method prior to the removal of the bin file I think though.  Which could be done my way or the puppetlabs-transition module way.  
I'm currently pushing for puppetlabs to consider including my code which i've put into a module for them to review as it works in Puppet 3 and 4 to create a fact and function that I think would be useful for modules like consul.
https://github.com/sfu-rcg/puppet-sfu_stdlib

We just had an issue where consul dumped on our network and I think it's due to servers lacking quorum with puppet managing their restart and adding 200+ clients.

aj-jester: Yup. When upgrading consul on a server it requires manual intervention. I usually stop puppet on *all* masters and then run puppet on each. I'll write a fix to optionally allow us to toggle service restarts so we can disable them on master so we don't have to keep stopping puppet.

asasfu: Slightly confused, you stop puppet on all consul servers and then run puppet one at a time? or run consul one at a time.  It seems to me like we should suggest to people that for consul servers to set manage_service to false and manually handle it for the time being.  That way if they update config_hash it doesn't cause a bootstrap quorum issue.

aj-jester: The problem with setting `manage_service` to false is during automation it won't bring up consul, which is not ideal. What we need to do implement a new feature called "restart on config change" where we can toggle whether to restart consul on config change. This way we can disable restart on consul servers but the consul will still start up on a newly provisioned node.

aj-jester: > I'll write a fix to optionally allow us to toggle service restarts so we can disable them on master so we don't have to keep stopping puppet.

https://github.com/solarkennedy/puppet-consul/pull/169

PR #169, opened 2015-08-14
parameterize restart on change for the main config

aj-jester:
This will allow us to restart the agent conditionally when `$config_hash` changes. This does not affect reloads when service, check and watch configs change.

It defaults to the current behavior of restarting on config change.

In my experience its never a good idea for puppet to restart consul `server`. More often than not you will run into a situation where the cluster doesn't know who the leader is. Its best to restart the agent on server in a more controlled fashion.


solarkennedy:
Thanks!

PR #171, opened 2015-08-14
deep_merge to support nested objects

aj-jester:
Updated to use `deep_merge` to support nested hashes.

Consul has 2 nested objects, [addresses](https://www.consul.io/docs/agent/options.html#addresses) and [ports](https://www.consul.io/docs/agent/options.html#ports) and the `merge` function doesn't properly merge nested hashes.


solarkennedy:
Thanks!

PR #172, opened 2015-08-14
Add a check for $service_ensure in reload_service

pforman:
When pre-provisioning a node with a tool like Packer (https://packer.io)
it's possible that we're only installing using Puppet, not attempting to
run a live configuration.  "reload_service" didn't respect the setting
of $service_ensure, so it would attempt to reload the consul config even
if the service was supposed to be stopped.

In a pre-provisioning case, the Consul config may be populated with
place-holders for values like keys or servers that are expected to be
replaced at launch time (example: AWS EC2 user-data).

The "reload_service" class is a notification dependency for many things,
so disabling it at each point is difficult and error-prone.  This commit
has "reload_service" ensure that the value of $service_ensure is not
'stopped', and only performs the reload if it's appropriate.

Otherwise, the method is a no-op.

Here's the relevant snippet of my output when running with server_ensure => 'stopped' before the patch.  This shows Packer's masterless Puppet provisioner failing to reload the Consul config, because it has placeholder syntax errors that user-data will fix on a real instance:

    amazon-ebs: Info: Class[Consul::Config]: Scheduling refresh of Class[Consul::Run_service]
    amazon-ebs: Info: Class[Consul::Run_service]: Scheduling refresh of Service[consul]
    amazon-ebs: Notice: /Stage[main]/Consul::Run_service/Service[consul]/enable: enable changed 'false' to 'true'
    amazon-ebs: Notice: /Stage[main]/Consul::Run_service/Service[consul]: Triggered 'refresh' from 1 events
    amazon-ebs: Info: Class[Consul::Reload_service]: Scheduling refresh of Exec[reload consul service]
    amazon-ebs: Notice: /Stage[main]/Consul::Reload_service/Exec[reload consul service]/returns: Error connecting to Consul agent: dial tcp 127.0.0.1:8400: connection refused
    amazon-ebs: Error: /Stage[main]/Consul::Reload_service/Exec[reload consul service]: Failed to call refresh: consul reload returned 1 instead of one of [0]
    amazon-ebs: Error: /Stage[main]/Consul::Reload_service/Exec[reload consul service]: consul reload returned 1 instead of one of [0]
    amazon-ebs: Info: Creating state file /var/lib/puppet/state/state.yaml
    amazon-ebs: Notice: Finished catalog run in 18.57 seconds
==> amazon-ebs: Terminating the source AWS instance...

Packer fails the build if any of the tools return an error, so the reload failure trickles up and aborts the AMI creation.

Here's a post-patch snippet.  "reload_service" is called, but becomes effectively a no-op.  The setup completes happily.

    amazon-ebs: Info: Class[Consul::Config]: Scheduling refresh of Class[Consul::Run_service]
    amazon-ebs: Info: Class[Consul::Run_service]: Scheduling refresh of Service[consul]
    amazon-ebs: Notice: /Stage[main]/Consul::Run_service/Service[consul]/enable: enable changed 'false' to 'true'
    amazon-ebs: Notice: /Stage[main]/Consul::Run_service/Service[consul]: Triggered 'refresh' from 1 events
    amazon-ebs: Info: Creating state file /var/lib/puppet/state/state.yaml
    amazon-ebs: Notice: Finished catalog run in 17.86 seconds
==> amazon-ebs: Stopping the source instance...

I realize this may be a bit of an odd case, and most people don't run with enable => true AND ensure => 'stopped', but it presents a real problem for our pre-built AMI workflow.

I'm totally open to discussing other ways to accomplish this as well.

### DIFF
--- a/manifests/reload_service.pp
+++ b/manifests/reload_service.pp
@@ -9,7 +9,7 @@ class consul::reload_service {
 
   # Don't attempt to reload if we're not supposed to be running.
   # This can happen during pre-provisioning of a node.
-  if $::consul::service_ensure != 'stopped' {
+  if $consul::manage_service == true and $consul::service_ensure == 'running' {
     exec { 'reload consul service':
       path        => [$consul::bin_dir,'/bin','/usr/bin'],
       command     => 'consul reload',
--- a/manifests/reload_service.pp
+++ b/manifests/reload_service.pp
@@ -7,10 +7,14 @@
 #
 class consul::reload_service {
 
-  exec { 'reload consul service':
-    path        => [$consul::bin_dir,'/bin','/usr/bin'],
-    command     => 'consul reload',
-    refreshonly => true,
+  # Don't attempt to reload if we're not supposed to be running.
+  # This can happen during pre-provisioning of a node.
+  if $::consul::service_ensure != 'stopped' {
+    exec { 'reload consul service':
+      path        => [$consul::bin_dir,'/bin','/usr/bin'],
+      command     => 'consul reload',
+      refreshonly => true,
+    }
   }
 
 }
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -270,6 +270,30 @@ describe 'consul' do
     it { should_not contain_service('consul') }
   end
 
+  context "When a reload_service is triggered with service_ensure stopped" do
+    let (:params) {{
+      :service_ensure => 'stopped',
+      :services => {
+        'test_service1' => {
+          'port' => '5'
+        }
+      }
+    }}
+    it { should_not contain_exec('reload consul service')  }
+  end
+
+  context "When a reload_service is triggered with manage_service false" do
+    let (:params) {{
+      :manage_service => false,
+      :services => {
+        'test_service1' => {
+          'port' => '5'
+        }
+      }
+    }}
+    it { should_not contain_exec('reload consul service')  }
+  end
+
   context "With a custom username" do
     let(:params) {{
       :user => 'custom_consul_user',
@@ -291,6 +315,7 @@ describe 'consul' do
     }}
     it { should contain_consul__service('test_service1').with_port('5') }
     it { should have_consul__service_resource_count(1) }
+    it { should contain_exec('reload consul service')  }
   end
 
   context "When the user provides a hash of watches" do
@@ -305,6 +330,7 @@ describe 'consul' do
     it { should contain_consul__watch('test_watch1').with_type('nodes') }
     it { should contain_consul__watch('test_watch1').with_handler('test.sh') }
     it { should have_consul__watch_resource_count(1) }
+    it { should contain_exec('reload consul service')  }
   end
 
   context "When the user provides a hash of checks" do
@@ -319,6 +345,7 @@ describe 'consul' do
     it { should contain_consul__check('test_check1').with_interval('30') }
     it { should contain_consul__check('test_check1').with_script('test.sh') }
     it { should have_consul__check_resource_count(1) }
+    it { should contain_exec('reload consul service')  }
   end
 
   context "With multiple watches and a config hash for #83" do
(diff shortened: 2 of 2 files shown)


manifests/reload_service.pp
```diff
@@ -7,10 +7,14 @@
 #
 class consul::reload_service {
 
-  exec { 'reload consul service':
-    path        => [$consul::bin_dir,'/bin','/usr/bin'],
-    command     => 'consul reload',
-    refreshonly => true,
+  # Don't attempt to reload if we're not supposed to be running.
+  # This can happen during pre-provisioning of a node.
+  if $::consul::service_ensure != 'stopped' {
```
aj-jester: I think we should make this a little more robust:

```
if $consul::manage_service == true and $consul::service_enable == true and $consul::service_ensure == 'running' {
 exec {...}
}
```

Also no need to prefix the variable with `::`, it can just be `$consul::some_param`

aj-jester:
@pforman can you please add some tests as well for this?

pforman:
Whoops, forgot about the $manage_service param.  Good catch.  If Puppet isn't managing the service at all it really shouldn't try to reload.

I think $service_enable shouldn't be part of that check, because it would break the very workflow that I was trying to fix.  In my case, I'm building images that make use of Consul on startup, so I want "enable => true" to set up start-on-boot, but I don't want the service to be running during the provisioning process.  Seems that $service_ensure and $manage_service cover it sufficiently.

I'll adjust the PR, look into some tests and resubmit.  Thanks for the quick response.

aj-jester:
@pforman Ok, thats fine. I was iffy about adding `$service_enable` part of the conditional but after your reasoning I think it makes sense to leave it out. Thanks.

pforman:
Updated and added tests.

Tests were a bit weird, because reload_service doesn't have any parameters.   I used the same methods as the run_service tests and put these in init_spec.pp

I also noticed the existing service/check hash tests (in init_spec.pp, not the defines) weren't confirming the reload should happen, so I added that confirmation.

All passes on my end.

aj-jester:
@pforman nice work on this :tada: 

:bullettrain_side: LGTM

solarkennedy:
Thanks!

CruzanCaramele:
hi @solarkennedy @pforman , I just cloned the repo and trying to provision an AMI using packer. However am getting a similar error as below:

```
amazon-ebs: Error: Could not start Service[consul]: Execution of '/sbin/start consul' returned 1: start: Job failed to start
    amazon-ebs: Error: Could not start Service[consul]: Execution of '/sbin/start consul' returned 1: start: Job failed to start
2016/02/07 22:17:57 ui:     amazon-ebs: Error: /Stage[main]/Consul::Run_service/Service[consul]/ensure: change from stopped to running failed: Could not start Service[consul]: Execution of '/sbin/start consul' returned 1: start: Job failed to start
    amazon-ebs: Error: /Stage[main]/Consul::Run_service/Service[consul]/ensure: change from stopped to running failed: Could not start Service[consul]: Execution of '/sbin/start consul' returned 1: start: Job failed to start
```
 Please how may i fix this ?

aj-jester:
Packer by default doesn't have `/sbin` in the `$PATH` variable. Internally `start`, `stop`, `restart` etc. are basically bin scripts under `/sbin`.

bflad:
Are there any logs in `/var/log/upstart/consul.log`? They might indicate any configuration errors of Consul itself, which would prevent startup

CruzanCaramele:
@bflad am not able to check `/var/log/upstart/consul.log` because the AMI does not get created by Packer due to the error.

bflad:
You might be able to catch it with the Packer `-debug` flag, but depends on the Packer provisioner implementation. I've never tried it myself on AWS. See https://github.com/mitchellh/packer/issues/409 and the current documentation for Debugging Packer Builds for how to SSH into your instances while they are building: https://www.packer.io/docs/other/debugging.html

CruzanCaramele:
I found the problem, sorry, this was mt mistake, all i had to do was add ** manage_service => false** which gets rid of the error when baking AMI with packer

yizhilong28:
+1

PR #173, opened 2015-08-19
prevent unnecessary consul restarts on puppet runs

mdelagrange:
if you are running consul RPC on a custom port, the startup script can now see that the service is running

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -115,7 +115,7 @@ class consul (
   if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
     $rpc_port = $config_hash_real['ports']['rpc']
   } else {
-    $rpc_port = '8400'
+    $rpc_port = 8400
   }
 
   if ($config_hash_real['addresses'] and $config_hash_real['addresses']['rpc']) {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -113,7 +113,17 @@ class consul (
   }
 
   if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
-    $custom_rpc_port = $config_hash_real['ports']['rpc']
+    $rpc_port = $config_hash_real['ports']['rpc']
+  } else {
+    $rpc_port = '8400'
+  }
+
+  if ($config_hash_real['addresses'] and $config_hash_real['addresses']['rpc']) {
+    $rpc_addr = $config_hash_real['addresses']['rpc']
+  } elsif ($config_hash_real['client_addr']) {
+    $rpc_addr = $config_hash_real['client_addr']
+  } else {
+    $rpc_addr = $::ipaddress_lo
   }
 
   if $services {
@@ -144,7 +154,6 @@ class consul (
     config_hash => $config_hash_real,
     purge       => $purge_config_dir,
     notify      => $notify_service,
-    rpc_port    => $custom_rpc_port,
   } ->
   class { 'consul::run_service': } ->
   class { 'consul::reload_service': } ->
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -112,6 +112,10 @@ class consul (
     warning('data_dir must be set to install consul web ui')
   }
 
+  if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
+    $custom_rpc_port = $config_hash_real['ports']['rpc']
+  }
+
   if $services {
     create_resources(consul::service, $services)
   }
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -140,6 +144,7 @@
     config_hash => $config_hash_real,
     purge       => $purge_config_dir,
     notify      => $notify_service,
+    rpc_port    => $custom_rpc_port,
```
aj-jester: I don't think you need to pass this as a param, you can just do:

```
if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
  $rpc_port = $config_hash_real['ports']['rpc']
} else {
  $rpc_port = 8400
}
```

then while calling it in `config.pp` just do `$consul::rpc_port`

manifests/init.pp
```diff
@@ -112,6 +112,10 @@
     warning('data_dir must be set to install consul web ui')
   }
 
+  if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
+    $custom_rpc_port = $config_hash_real['ports']['rpc']
+  }
```
aj-jester: it a good practice to initialize all variables, even if you have to do `undef`. In other words it should be like:

```
} else {
  $custom_rpc_port = 8400
}
```

Also, can we just name this `rpc_port`?

manifests/init.pp
```diff
@@ -112,6 +112,10 @@
     warning('data_dir must be set to install consul web ui')
   }
 
+  if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
+    $custom_rpc_port = $config_hash_real['ports']['rpc']
```
aj-jester: I would suggest renaming this to `$rpc_port`

templates/consul.sysconfig.erb
```diff
@@ -0,0 +1 @@
+RPC_ADDR=<%= "-rpc-addr=%{::ipaddress_lo}:#{@rpc_port}" -%>
```
aj-jester: `::ipaddress_lo` is a concern because you can configure `rpc` address in consul as well.

So in `init.pp` you need another conditional:
```
if $config_hash_real['rpc'] {
  $rpc_addr = $config_hash_real['rpc']
} elsif $config_hash_real['client_addr'] {
  $rpc_addr = $config_hash_real['client_addr']
} else {
  $rpc_addr = '127.0.0.1'
}
```

Then remove this file entirely `consul.sysconfig.erb` and use these params directly in `consul.sysv.erb like so:
`"$CONSUL" leave <%= @rpc_addr %>:<%= @rpc_port %>`

This way we can use this with other init scripts as well.

https://www.consul.io/docs/agent/options.html#rpc

aj-jester:
@mdelagrange It would be awesome if there are some tests to go along with this since there are a lot of moving parts to this.

mdelagrange:
@aj-jester yep that all sounds good.  i'll get to work sharpish.

mdelagrange:
@aj-jester updated per your recommendations.  thanks for the feedback!

aj-jester:
@mdelagrange This looks stupendous! Nice work dude :100:

:bullettrain_side: LGTM

manifests/init.pp
```diff
@@ -112,6 +112,20 @@
     warning('data_dir must be set to install consul web ui')
   }
 
+  if ($config_hash_real['ports'] and $config_hash_real['ports']['rpc']) {
+    $rpc_port = $config_hash_real['ports']['rpc']
+  } else {
+    $rpc_port = '8400'
```
aj-jester: oh! I missed this, no quoting integers please :)

mdelagrange:
Fixed.

aj-jester:
Can we hold up on merging this? I just want to make sure we don't commit just to revert all this back because of https://github.com/solarkennedy/puppet-consul/issues/174

mdelagrange:
To me it sounds like @pforman was waiting for this to be merged before he does his PR. Regardless of the solution to #174 it's going to need this logic to work properly on a custom host or port.

aj-jester:
We want to understand how consul behaves if `leave_on_terminate` is set to `true` and `TERM` signal is sent with `consul leave` executing right before the `TERM`.

pforman:
I don't have any issue with this, and it won't greatly affect anything I've been talking about.  As @mdelagrange says, the RPC info is going to need to be there anyway.  If it's merged, I'll do my TERM-based PR from the new HEAD.  If we do them at the same time, it would be a definite conflict, as we're changing the very same lines in the sysv script.

The only thing it would change for me is the addition of the ${RPC_ADDR}, which I'd preserve.

aj-jester:
ok this should be good to merge :+1:

solarkennedy:
Thanks! I'll work on porting this feature to the other init systems.

aj-jester:
@solarkennedy oh thats awesome! It was on my list but now I can cross it off :stuck_out_tongue:

Thanks.

PR #176, opened 2015-08-22
Archlinux support

vdloo:
seems to work without any other modifications
```
Notice: /Stage[main]/Consul::Run_service/Service[consul]/ensure: ensure changed 'stopped' to 'running'
```

Archlinux only has official support for systemd

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -52,6 +52,8 @@ class consul::params {
     } else {
       $init_style = 'systemd'
     }
+  } elsif $::operatingsystem == 'Archlinux' {
+    $init_style = 'systemd'
   } elsif $::operatingsystem == 'SLES' {
     $init_style = 'sles'
   } elsif $::operatingsystem == 'Darwin' {
--- a/metadata.json
+++ b/metadata.json
@@ -61,6 +61,9 @@
       ]
     },
     {
+      "operatingsystem": "Archlinux"
+    },
+    {
       "operatingsystem": "Fedora",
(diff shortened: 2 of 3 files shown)


aj-jester:
@vdloo This is great :tada: 

But a few things...

* Can you add the [supported versions](https://github.com/solarkennedy/puppet-consul/blob/master/metadata.json#L21) of arch linux? It doesn't have to be anything exhaustive but what you know.
* Can we write a unit test for this? Here is an [example](https://github.com/solarkennedy/puppet-consul/blob/master/spec/classes/init_spec.rb#L417)

Thanks :smile:

vdloo:
Sure, but because of arch's [rolling release system](https://en.wikipedia.org/wiki/Arch_Linux#Versions) there aren't really versions like in Debian or RedHat though. I put in the 'operatingsystemrelease' that facter returned on my system but that's just the kernel version and can change between [releases](https://www.archlinux.org/releng/releases/). So if the metadata.json is going to stay current it's probably going to need to be changed a [couple times a month](https://projects.archlinux.org/svntogit/packages.git/log/trunk?h=packages/linux).
```
# facter operatingsystemrelease
4.1.5-1-ARCH
```

aj-jester:
@vdloo actually, lets do something like [this](https://github.com/garethr/garethr-docker/blob/master/metadata.json#L29).

aj-jester:
:100: 

:bullettrain_side: LGTM

solarkennedy:
Thanks!

PR #177, opened 2015-08-23
More rpc port support for debian/upstart

solarkennedy:
Related to #173 

I wish we didn't have so many init scripts to maintain :(

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -448,7 +448,40 @@ describe 'consul' do
     }
   end
 
-   context "When using upstart" do
+   context "When using debian" do
+    let (:params) {{
+      :init_style => 'debian'
+    }}
+    let (:facts) {{
+      :ipaddress_lo => '127.0.0.1'
+    }}
+    it { should contain_class('consul').with_init_style('debian') }
+    it {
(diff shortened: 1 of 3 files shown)


aj-jester:
I tested it on Ubuntu and it works :+1: 

One trivial suggestion I would make is adding a test when the `rpc_addr` == `client_addr` because of this https://github.com/solarkennedy/puppet-consul/blob/master/manifests/init.pp#L133

So something like this:

```
ontext "When rpc_addr defaults to client_addr on sysv" do
    let (:params) {{
      :init_style => 'sysv',
      :config_hash => {
        'client_addr' => '192.168.34.56',
      }
    }}
    it { should contain_class('consul').with_init_style('sysv') }
    it {
      should contain_file('/etc/init.d/consul').
        with_content(/-rpc-addr=192.168.34.56:8400/)
    }
  end
```

:bullettrain_side: LGTM

aj-jester:
@solarkennedy Sorry this is my fault, I should've asked for this test to be added in the other PR as well, I didn't think about it then, so yah no big deal if we don't want to add.

solarkennedy:
Added! Thanks.

PR #179, opened 2015-08-24
fix 'consul reload' on custom rpc port

mdelagrange:
Similar to #173.

### DIFF
--- a/manifests/reload_service.pp
+++ b/manifests/reload_service.pp
@@ -13,6 +13,9 @@ class consul::reload_service {
     exec { 'reload consul service':
       path        => [$consul::bin_dir,'/bin','/usr/bin'],
       command     => 'consul reload',
+      environment => [
+        "CONSUL_RPC_ADDR=${consul::rpc_addr}:${consul::rpc_port}",
+      ],
       refreshonly => true,
     }
   }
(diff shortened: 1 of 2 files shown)


manifests/reload_service.pp
```diff
@@ -13,6 +13,9 @@
     exec { 'reload consul service':
       path        => [$consul::bin_dir,'/bin','/usr/bin'],
       command     => 'consul reload',
+      environment => [
+        "CONSUL_RPC_ADDR=${consul::rpc_addr}:${consul::rpc_port}",
```
aj-jester: can we just do this as part of the command?

```
command => "consul reload -rpc-addr=${consul::rpc_addr}:${consul::rpc_port}",
```

mdelagrange:
@ericpfisher pointed out that, if we used the environment variable rather than the flag, we could probably make the init scripts cleaner.

aj-jester:
@mdelagrange Can we add a test for the `elsif`? https://github.com/solarkennedy/puppet-consul/blob/master/manifests/init.pp#L133

Also it would be awesome, but not required, if we can add a test for the `sysv` (your previous PR) for this `elsif` conditional as well. I'll do the same for the `upstart` :smile:.

> if we used the environment variable rather than the flag, we could probably make the init scripts cleaner.

Hmmm perhaps but I prefer if we are more explicit with the flag :confused:

mdelagrange:
OK

mdelagrange:
All set.

aj-jester:
@mdelagrange this looks great :+1: 

:bullettrain_side: LGTM

solarkennedy:
Thanks!

Issue #174, opened 2015-08-20
Strange (probably unnecessary) behavior in sysv stop script


pforman:
The sysv stop script has this in it for the stop section:

```
stop() {
        echo -n "Shutting down consul: "
        # If consul is not acting as a server, exit gracefully
        if ("${CONSUL}" info 2>/dev/null | grep -q 'server = false' 2>/dev/null) ; then
            "$CONSUL" leave
        fi

        # If acting as a server, or if leave failed, kill it.
        mkpidfile
        killproc $KILLPROC_OPT $CONSUL -9

        retcode=$?
        rm -f /var/lock/subsys/consul $PID_FILE
        return $retcode
}
```

I read the original PR (#87), and I agree with the theory to have clients leave but servers stay in "failed" state to preserve their state for a rejoin.  However, the implementation doesn't seem to address this correctly, and "kill -9" is extremely heavy-handed for a distributed consensus system.  It doesn't seem like that's ever going to be the right move.

There's also a problem where if the leave works on a client, the kill will fail, resulting in a "FAILED" response from $retcode.

I also have observed some cases of clients in a "failed" state where they should have left, which I think is down to a race condition between issuing a leave and the subsequent 'kill -9'.

I have a PR almost ready to go for this, but then I saw #173 in the queue working in exactly the same files (and same lines), so to avoid a conflict I've held off.

I also figured some discussion about what the actual effect should be was in order.  Consul will normally quit quickly (without issuing a leave) when given a TERM, however this can be controlled by the leave_on_terminate config option.  Seems like issuing a TERM is correct for servers wanting to preserve state, and can still be controlled if desired in the config_hash.

What to do with a client that fails to leave is a little harder.  In a few cases, I've seen a failure to leave immediately, which manifested as this message.

> Shutting down consul: Error leaving: client closed

However, looking in the logs it appears this is a temporary issue in resending gossip, and doesn't actually affect the leave process.

The logic I've used is like this (irrelevant stuff removed):

```
  if client
     leave OR kill -TERM
     retcode = $?
  else
    kill -TERM
    retcode = $?
  end
  return retcode $?
```

But honestly, I'm questioning the use of the TERM case at all in the client section.  Any thoughts on this before I send in a PR?

As far as I can tell, this "kill -9" usage is unique to the sysv script.  Every other method uses "consul" leave, or possibly TERM.  Debian escalates TERM to KILL after a timeout, but doesn't start there.

[...]

PR #181, opened 2015-08-26
Rework sysv stop script to fix issues

pforman:
Addresses #174.

I commented the stop function extensively so the next person in there will understand what I was trying to do and the signal logic there.  It's pretty similar to what it was doing before, but without the race condition and using more appropriate signals.

In my tests the SIGINT-triggered leave happens in less than 1 second, but takes enough time that a wait check is necessary to avoid the race.  We wait up to 5 seconds, checking every second to see if it's finished.

Commit message follows.

--
The sysv stop script had been set to issue a SIGKILL (-9) to ensure that
a server would not perform a leave operation.  That's excessive, and the
logic in that section had a race condition that could result in a client
unpredictably ending in either a "left" or "failed" state.

Reworked this section to issue a SIGINT to clients (which initiates a
leave by default) and then check for success.  If the leave operation
fails or if the agent is acting as a server, then the standard killproc
is issued, which tries a SIGTERM, followed by a SIGKILL if the TERM is
not effective.

Also fixed a minor issue where the stop return status would often be
displayed in the prompt line.  The start status is eaten by Consul's
default stdout logging, so there's no fixing that.  It simply ends up in
the log file.

### DIFF
--- a/templates/consul.sysv.erb
+++ b/templates/consul.sysv.erb
@@ -96,7 +96,7 @@ stop() {
         killproc $KILLPROC_OPT $CONSUL
         retcode=$?
 
-        rm -f /var/lock/subsys/consul
+        rm -f /var/lock/subsys/consul $PID_FILE
         return $retcode
 }
 
--- a/templates/consul.sysv.erb
+++ b/templates/consul.sysv.erb
@@ -61,18 +61,42 @@ start() {
 }
 
 stop() {
+        DELAY=5 # seconds maximum to wait for a leave
+
         echo -n "Shutting down consul: "
+        mkpidfile
+
         # If consul is not acting as a server, exit gracefully
+        # Use SIGINT to create a "leave" event, unless the user has explicitly
+        # changed that behavior in the Consul config.
         if ("${CONSUL}" info ${RPC_ADDR} 2>/dev/null | grep -q 'server = false' 2>/dev/null) ; then
-            "$CONSUL" leave ${RPC_ADDR}
+            consul_pid=$(cat $PID_FILE)
+            killproc $KILLPROC_OPT $CONSUL -INT
+            retcode=$?
+
+            # We'll wait if necessary to make sure the leave works, and return
+            # early if we can.  If not, escalate to harsher signals.
+            try=0
+            while [ $try -lt $DELAY ]; do
+                if ! checkpid $consul_pid ; then
+                    rm -f /var/lock/subsys/consul
+                    return $retcode
+                fi
+                sleep 1
+                let try+=1
+            done
         fi
 
-        # If acting as a server, or if leave failed, kill it.
-        mkpidfile
-        killproc $KILLPROC_OPT $CONSUL -9
-
+        # If acting as a server, use a SIGTERM to avoid a leave.
+        # This behavior is also configurable.  Avoid doing a "leave" because
+        # having servers missing is a bad thing that we want to notice.
+        #
+        # A SIGTERM will mark the node as "failed" until it rejoins.
+        # killproc with no arguments uses TERM, then escalates to KILL.
+        killproc $KILLPROC_OPT $CONSUL
         retcode=$?
-        rm -f /var/lock/subsys/consul $PID_FILE
+
+        rm -f /var/lock/subsys/consul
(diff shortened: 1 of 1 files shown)


templates/consul.sysv.erb
```diff
@@ -61,18 +61,42 @@ start() {
...
+            done
+        fi
 
+        # If acting as a server, use a SIGTERM to avoid a leave.
+        # This behavior is also configurable.  Avoid doing a "leave" because
+        # having servers missing is a bad thing that we want to notice.
+        #
+        # A SIGTERM will mark the node as "failed" until it rejoins.
+        # killproc with no arguments uses TERM, then escalates to KILL.
+        killproc $KILLPROC_OPT $CONSUL
```
aj-jester: I realize you mentioned this in the comment but can we add `-TERM` anyway just to be explicit?

templates/consul.sysv.erb
```diff
@@ -61,18 +61,42 @@ start() {
...
+            killproc $KILLPROC_OPT $CONSUL -INT
+            retcode=$?
 
-        # If acting as a server, or if leave failed, kill it.
-        mkpidfile
-        killproc $KILLPROC_OPT $CONSUL -9
+            # We'll wait if necessary to make sure the leave works, and return
+            # early if we can.  If not, escalate to harsher signals.
+            try=0
+            while [ $try -lt $DELAY ]; do
```
aj-jester: One thing we need to consider is what would happen if the PID never dies and `$try` is greater than or equal to `$DELAY`. At that point it will break out of the loop and execute `-TERM`. I don't think thats desired?

templates/consul.sysv.erb
```diff
@@ -61,18 +61,42 @@ start() {
...
+        # This behavior is also configurable.  Avoid doing a "leave" because
+        # having servers missing is a bad thing that we want to notice.
+        #
+        # A SIGTERM will mark the node as "failed" until it rejoins.
+        # killproc with no arguments uses TERM, then escalates to KILL.
+        killproc $KILLPROC_OPT $CONSUL
         retcode=$?
-        rm -f /var/lock/subsys/consul $PID_FILE
+
+        rm -f /var/lock/subsys/consul
```
aj-jester: We are not deleting the PID file anymore?

pforman:
I'll try to roll up all the questions...  Digging into how "killproc" works, it issues a TERM and then spin-waits until the process shuts down or exceeds its delay (3s by default).  At that point it issues a KILL.

However, it *doesn't* do that if a specific signal is passed.  It merely issues that signal once and then hopes for the best.  That's why I didn't include -TERM on the server killproc invocation, because then it won't escalate to KILL.

I dropped the "else" clause on the client to take advantage of the signal escalation as well.  INT, wait, TERM, wait, KILL.   

If a leave fails, I think it's reasonable to escalate to a TERM, since something's clearly wrong with the node.  (I've never seen a leave fail, I'm not sure what circumstances could cause that, but perhaps a network partition?)

The $PID_FILE removal is normally handled by Consul, but in the case of a SIGKILL, that wouldn't happen.  I think that should be reverted to
```
rm -f /var/lock/subsys/consul $PID_FILE
````
Good catch, I'll admit it was getting late last night. 

I originally dropped that because the race to kill consul was removing it.  Then consul itself would complain it couldn't remove the file, because the `rm` had been there first.  Now that the wait happens in both client and server cases, it shouldn't cause any issues, and would be important in a SIGKILL case.

aj-jester:
> it was getting late last night

@pforman haha I know that feeling

So yah, I agree; INIT, wait, KILL and end early on success and clean up and any lock/pid files :+1:

pforman:
Okay, I'll test and submit an update to revert the $PID_FILE cleanup.  Thanks for the careful review.

solarkennedy:
Yuck I hate init scripts :(

I'll review this carefully this weekend.

solarkennedy:
Let's see how it goes!

pforman:
Thanks.  I tested it in my use cases (Amazon Linux, so RH-derived) pretty aggressively, including looping stop-start for quite a while.  I'll be embarrassed if I broke something :)

PR #183, opened 2015-09-10
Fixed chuid / consul executed as root

sw0x2A:
consul was running as root because -u was used (matching) but you want -c (chuid).


aj-jester:
@solarkennedy huh...the beaker tests are failing on `master` branch as well.

solarkennedy:
Hmm. I've disabled them for now. It looks like there are better ways to do this now (example: https://github.com/camptocamp/puppet-openldap/blob/master/.travis.yml)

solarkennedy:
Ty @sw0x2A ! I'm surprised this was missed...

PR #185, opened 2015-09-14
Install unzips conditionally

bearrito:
A fix for the following issue we have encountered
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: Package[unzip] is already declared; cannot redeclare at /etc/puppet/modules/consul/manifests/install.pp:19
```

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -16,8 +16,8 @@ class consul::install {
   case $consul::install_method {
     'url': {
       if $::operatingsystem != 'darwin' {
-        if ! defined ( Package['unzip'] ) { 
-          ensure_packages(['unzip'], { 'before' => Staging::File['consul.zip'] }) 
+        if ! defined ( Package['unzip'] ) {
+          ensure_packages(['unzip'], { 'before' => Staging::File['consul.zip'] })
         }
       }
       staging::file { 'consul.zip':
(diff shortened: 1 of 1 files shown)


solarkennedy:
Yea, puppet and parse order dependent code kinda sucks. I'm guessing you have a package { 'unzip': } somewhere else? (that isn't using ensure_packages)

Looks like there is some trailing whitespace, can you fix the tests?
https://travis-ci.org/solarkennedy/puppet-consul/jobs/80312563

bearrito:
I don't have enough ruby/puppet knowledge to fix that failing spec. Any advice on how to proceed?

solarkennedy:
In general:
https://travis-ci.org/solarkennedy/puppet-consul/jobs/80437216#L160

We tried to ensure that unzip comes before we try to unzip the web file, but it looks like it looks like your code introduced a bug:
https://github.com/solarkennedy/puppet-consul/pull/185/files#diff-a276806bea2dfcf1f57c33aee19c7309R39

Be sure that unzip comes before the normal consul.zip and consul_web

manifests/install.pp
```diff
@@ -16,7 +16,9 @@
   case $consul::install_method {
     'url': {
       if $::operatingsystem != 'darwin' {
-        ensure_packages(['unzip'], { 'before' => Staging::File['consul.zip'] })
+        if ! defined ( Package['unzip'] ) {
```
aj-jester: can you also please remove these additional white spaces? Something like this is preferable: `if !defined(Package['unzip']) {`

solarkennedy:
Closing as we are no longer managing the zip package.

PR #186, opened 2015-09-15
actually we want to escape / globally for filenames

duritong:
Introduce the same for services


aj-jester:
:bullettrain_side: LGTM

solarkennedy:
Thanks!

PR #188, opened 2015-09-21
consul init sysv: lower stop priority

koendc:
Lower the stop priority in order to make sure that a consul agent is
able to leave the cluster gracefully when shut down.


solarkennedy:
Thanks!

PR #189, opened 2015-09-23
Remove management of unzip package

danieldreier:
The unzip package is already managed by most people's puppet infrastructure, so installing it here introduces resource conflicts.

Even with the ensure_resource function, there are parse-order dependent conflicts if you have the same packages declared elsewhere.

I made this change because I get a resource conflict without it. I'd be open to another approach but wanted to contribute the change back upstream in case other people were running into the same problem.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -15,9 +15,6 @@ class consul::install {
 
   case $consul::install_method {
     'url': {
-      if $::operatingsystem != 'darwin' {
-        ensure_packages(['unzip'], { 'before' => Staging::File['consul.zip'] })
-      }
       staging::file { 'consul.zip':
         source => $consul::real_download_url
       } ->
@@ -32,9 +29,6 @@ class consul::install {
       }
 
       if ($consul::ui_dir and $consul::data_dir) {
-        if $::operatingsystem != 'darwin' {
-          Package['unzip'] -> Staging::Deploy['consul_web_ui.zip']
-        }
         file { "${consul::data_dir}/${consul::version}_web_ui":
           ensure => 'directory',
           owner  => 'root',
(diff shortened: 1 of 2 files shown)


solarkennedy:
I agree that ensure_resources isn't great. Other solutions for this are #185 and #123.

/me sighs....

@aj-jester what do you think? Even with other solutions where we do `! defined` we still can get conflicts with other modules declaring unzip *after* consul is included. Should we leave this just up as an exercise to the reader and document it?

split3:
This is one of those annoying situations with puppet and causes constant issues.  Not sure there is really a great workaround for this.  Another possibility solution would be (and I'm not a huge fan) but to add an option flag to on wether to manage that package or not, as well not ideal but workable.

solarkennedy:
Alright, I'm just going to make the call and say this is out of scope of the module, there have been enough pull requests to adjust this. I'll merge this and adjust the docs.

danieldreier:
thanks @solarkennedy - one solution I've seen elsewhere is to provide a sample profile that includes extra packages, in a namespace like consul::profile::server / consul::profile::agent. That way somebody who doesn't really know what they're doing can include it verbatim, and somebody who wants more control can copy parts into their own site control repo.

aj-jester:
I looked through `nanliu/staging` module and it looks like they are assuming the `unzip` package would be installed separately as well.

@solarkennedy So I think we can take that same approach and just document the fact `unzip` needs to be installed separately and remove it from our module.

If we want, as an added step, we can do `require => Package['unzip']` but I would like to hear why this might be a bad idea, cause it sounds like a bad idea to me :stuck_out_tongue_closed_eyes:

PR #197, opened 2015-11-11
Add parameter for setting port to custom acl type

horsehay:
@solarkennedy,

I need to set the port for consul, but it's hard coded into the type provider. Added a parameter to the type and set the default to 8500.

Thanks


solarkennedy:
Thanks!

PR #198, opened 2015-12-04
Clean keys

split3:
Moving these out since we'll just put them in our own configuration class/module


split3:
Targeted wrong repository

PR #205, opened 2015-12-14
mitigate .to_json segfaults on Ruby 1.8.7

duritong:
Regularly we had processes on our puppetmaster - (still) running on
Ruby 1.8.7 - segfaulting on line 13 of consul_sorted_json.rb and we
then even had ghost processes (not tracked by passenger anymore)
consuming 100% of memory doing nothing when attaching with strace.

This might be a problem of either Ruby 1.8.7 itself (similar to
https://groups.google.com/forum/#!msg/puppet-users/ag1rAF7o_Uw/1hjtABrmzwgJ)
or the used C json library. Looking at what we are doing on line 13,
we figured out, that we could reduce the calls to .to_json and
hence might be able to mitigate the segfaults.

Only changing line 13 helped, but moved the segfaults to line 10,
looking at that line, we figured out, that we can avoid even
further .to_json calls.

Running this patch on 2 puppet masters for 8 hours didn't show any
segfaults, nor any high running cpu processes, which usually started
to appear quite quickly.

### DIFF
--- a/lib/puppet/parser/functions/consul_sorted_json.rb
+++ b/lib/puppet/parser/functions/consul_sorted_json.rb
@@ -6,11 +6,8 @@ module JSON
 
     def sorted_generate(obj)
       case obj
-        when Fixnum, Float, TrueClass, FalseClass, NilClass
-          return obj.to_json
-        when String
-          # Convert quoted integers (string) to int
-          return (obj.match(/\A[-]?[0-9]+\z/) ? obj.to_i : obj).to_json
+        when NilClass,Fixnum, Float, TrueClass, FalseClass,String
+          return simple_generate(obj)
         when Array
           arrayRet = []
           obj.each do |a|
@@ -34,14 +31,8 @@ module JSON
       indent = " " * indent_len
 
       case obj
-
-        when Fixnum, Float, TrueClass, FalseClass, NilClass
-          return obj.to_json
-
-        when String
-          # Convert quoted integers (string) to int
-          return (obj.match(/\A[-]?[0-9]+\z/) ? obj.to_i : obj).to_json
-
+        when NilClass,Fixnum, Float, TrueClass, FalseClass,String
+          return simple_generate(obj)
         when Array
           arrayRet = []
 
@@ -80,6 +71,20 @@ module JSON
       end
 
     end # end def
+    private
+    # simplify jsonification of standard types
+    def simple_generate(obj)
+      case obj
+        when NilClass
+          'null'
+        when Fixnum, Float, TrueClass, FalseClass
+          "#{obj}"
+        else
+          # Should be a string
+          # keep string integers unquoted
+          (obj =~ /\A[-]?\d+\z/) ? obj : obj.to_json
+      end
+    end
 
   end # end class
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_install: rm Gemfile.lock || true
 sudo: false
 cache: bundler
 rvm:
+  - 1.8.7
   - 1.9.3
   - 2.0.0
   - 2.1.0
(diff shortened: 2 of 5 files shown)


lib/puppet/parser/functions/consul_sorted_json.rb
```diff
@@ -6,11 +6,13 @@ class << self
 
     def sorted_generate(obj)
       case obj
-        when Fixnum, Float, TrueClass, FalseClass, NilClass
-          return obj.to_json
+        when NilClass
+          return "null"
```
solarkennedy: Uh, are you sure you want to literally return the string "null"? This sounds like a bad idea.

solarkennedy:
Keep in mind that we do not officially support old versions of ruby in this module (although it isn't super clear from the readme)

If this is a real issue for you, I encourage you to submit a PR that re-expands the scope of the .travis.yaml to include your use case.
This will not prevent me from accepting this pr, just an fyi

lib/puppet/parser/functions/consul_sorted_json.rb
split3: Yea I'd agree, returning the String value for a NilClass seems like a bad idea

aj-jester:
If I remember correctly the last time we added `1.8.7` support it broke tests and we ended up removing it https://github.com/solarkennedy/puppet-consul/pull/154

lib/puppet/parser/functions/consul_sorted_json.rb
duritong: Why?

    $ ruby -rjson -e 'puts nil.to_json'
    null

This is what .to_json gives back on nil-

solarkennedy: Well, yes when you print it, it becomes a string, yes.

I understand that this function actually generates "content" for puppet, so maybe returning the string is ok. 

Ug, I just don't like this function :(

Don't forget to adjust the "non"-pretty function as well.

Please add a test for this, I don't want to see this regress. 

Something like:

    it 'handles null correctly' do
      expect(subject.call({"foo" => null}).to eql('"foo"=>null')
    end

split3: I'd think that generating content having it return the value "null" becomes more problematic IMO.  I'd personally like if something is actually null the generated content is empty "".

duritong: Well, but that's not how the current implementation and also how json is specified.

Remember JSON is JavaScript Object Notation and in JavaScript nil is null.

I'm fine with adding the tests that represent/spec the current (pre-this-patch) behavior and hence I will show that the patch does not change it's current behvaior. I will also try to enable travis tests for 1.8.7, so that we could still support this ancient plattform. Yes, I'm very sad to not yet have had the option to move to puppet-server and hence jruby, BUT given that I will need this patch to have the current platform still working properly I thought it makes sense to give such things upstream, as others might run into the same problem on EL6.

split3: We just need to ensure that it is returning actual null, not the string "null" as those are in fact 2 different things in the JSON spec.

duritong: I agree, but then I would need to return "'null'" (quotes in the string), but we don't want that. `.to_json` returns a string that is then put into the template. It's confusing, we have layers of strings here...

duritong:
So I added:

* tests that show the current behavior of simple times *before* my main commit, hence I would say that I'm showing that I don't introduce any regression with my optimizations.
* put the optimizations into a private method, so that both - pretty and non pretty - generate methods can benefit from it.
* re-enabled 1.8.7 support and fixed 2 minor issues in the specs
* support for puppet 3.8.4
* fixed an old stubbing behavior that was showing a warning

If you think that anything is missing, please let me know.

solarkennedy:
Well. This looks pretty solid! Thanks!

Issue #204, opened 2015-12-10
Consul 0.6 support


tzz:
I noticed these were missing to support Consul 0.6:

* update download URLs and default version
* support `tcp` checks

Do you need patches for those?

split3:
The download url should be fixed already along with upgrading via url automatically via #201 and #202 

Haven't see anything for `tcp` checks, so patches welcome for that.

split3:
Also the default version wasn't adjust to 0.6.0 currently either.

tzz:
For the TCP checks, I applied this simple patch (also fixing a message typo). Feel free to use it; it works for me and it's a tiny change.

```
commit 874c1967993dc6da17881c92a0070e284439a660
Author: Ted Zlatanov <tzz@lifelogs.com>
Date:   Thu Dec 10 12:20:24 2015 -0500

    consul: allow tcp checks

diff --git a/modules/consul/lib/puppet/parser/functions/consul_validate_checks.rb b/modules/consul/lib/puppet/parser/functions/consul_validate_checks.rb
index 7fdd0eb..5772f38 100644
--- a/modules/consul/lib/puppet/parser/functions/consul_validate_checks.rb
+++ b/modules/consul/lib/puppet/parser/functions/consul_validate_checks.rb
@@ -15,7 +15,11 @@ def validate_checks(obj)
           end
         elsif obj.key?("http")
           if (! obj.key?("interval"))
-            raise Puppet::ParseError('http must be defined for interval checks')
+            raise Puppet::ParseError('interval must be defined for http checks')
+          end
+        elsif obj.key?("tcp")
+          if (! obj.key?("interval"))
+            raise Puppet::ParseError('interval must be defined for tcp checks')
           end
         elsif obj.key?("script")
           if (! obj.key?("interval"))
```

tzz:
For the URLs, the #201 and #202 fixes look great.

Changing the default version is up to you, of course.

PR #206, opened 2015-12-14
Added support for TCP checks (available in Consul 0.6.x)

split3:

PR #207, opened 2015-12-14
Add support for tcp check type

dbeckham:
These change build on @tzz's patch in #204, and adds full support for the `tcp` check type. I've updated both the `check.pp` manifest and the `consul_validate_checks()` parser function to properly support the type, as well as added tests that emulate the testing done for the other check types. If you need any changes made, or I missed something, let me know and I can update the branch.


split3:
Similar provided as well via #206, I'll merge these 2 together as you have some additional checks and docs.

split3:
@dbeckham I've merged your changes into #206 and we'll review and approve from that one.  I've preserved your commits so you will still get credit for the associated changes.

dbeckham:
Awesome! Thanks for accepting my changes.

PR #208, opened 2015-12-16
Fix web_ui installation on Consul 0.6.0 and greater

dbeckham:
The web_ui distribution file in Consul 0.6.0 no longer has a `dist/` directory as part of the archive's path. This change adds support web_ui archives for versions >= 0.6.0 as well as maintain compatibility with the older releases. It also fixes a specific issue where the generic `consul_web_ui.zip` file name would block upgrades to newer versions.  I've also updated existing tests, and written new tests for the version check.

If you think the version check in install.pp could be clearer, or if there is anything else you would prefer to be done differently, let me know and I'll be glad to update it.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,6 +1,6 @@
-# == Class consul::intall
+# == Class consul::install
 #
-# Installs consule based in the parameters from init
+# Installs consul based on the parameters from init
 #
 class consul::install {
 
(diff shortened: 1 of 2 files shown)


bjoernhaeuser:
Hit the same problem :-) Code looks good to me!

split3:
Thanks @dbeckham this looks pretty good to me! I know versioncmp can be flaky but I think it is the only real good way to do it, as that is what package uses under the hood now so hopefully it is more stable these days.

PR #209, opened 2015-12-17
Consul Version Fact

robrankin:
Bit rough and ready, improvements welcome.  Useful to see if all your servers have moved versions when performing an upgrade.

### DIFF
--- a/lib/facter/consul_version.rb
+++ b/lib/facter/consul_version.rb
@@ -3,7 +3,7 @@
 Facter.add(:consul_version) do
   confine :kernel => 'Linux'
   setcode do
-    version = Facter::Util::Resolution.exec('consul --version')
+    version = Facter::Util::Resolution.exec('consul --version 2> /dev/null')
     version = version.lines.first.split[1].tr('v','')
     version
   end
--- a/lib/facter/consul_version.rb
+++ b/lib/facter/consul_version.rb
@@ -10,6 +10,8 @@ end
 Facter.add(:consul_version) do
   confine :kernel => 'windows'
   setcode do
-    Facter::Core::Execution.exec('powershell (((& "C:\Program Files\Consul\consul.exe" "--version") | select -First 1).Split(" ",[StringSplitOptions]"RemoveEmptyEntries")[1]).TrimStart("v")')
+    powershell = 'powershell'
+    command = ' (((&\"C:\Program Files\Consul\consul.exe\" --version)[0]).Split(\"Consul v\")[8]).trim() '
+    Facter::Util::Resolution.exec(%Q{#{powershell} -command "#{command}"})
   end
 end
--- a/lib/facter/consul_version.rb
+++ b/lib/facter/consul_version.rb
@@ -1,15 +1,10 @@
 # consul_version.rb
 
 Facter.add(:consul_version) do
+  confine :kernel => 'Linux'
   setcode do
-    if Facter.value(:kernel) == 'Linux'
-      version = Facter::Util::Resolution.exec('consul --version')
-      version = version.lines.first.split[1].tr('v','')
-    elsif Facter.value(:kernel) == 'windows'
-      version = Facter::Util::Resolution.exec('"C:\\Program Files\\Consul\\consul.exe" --version')
-      version = version.lines.first.split[1].tr('v','')
-    end
-
+    version = Facter::Util::Resolution.exec('consul --version')
+    version = version.lines.first.split[1].tr('v','')
     version
   end
 end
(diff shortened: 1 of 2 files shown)


lib/facter/consul_version.rb
```diff
@@ -0,0 +1,17 @@
+# consul_version.rb
+
+Facter.add(:consul_version) do
+  confine :kernel => 'Linux'
+  setcode do
+    Facter::Core::Execution.exec('consul --version | head -1 | cut -f2 -d\ | cut -c2-')
```
solarkennedy: Can you do the raw exec and do the string manipulation in ruby? Same with powershell.

solarkennedy:
Add a test please for both linux and windows. Here is a similar example if you have not done so before:
https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/unit/facter/mysql_version_spec.rb

lib/facter/consul_version.rb
robrankin: That's a very good idea, can't believe I didn't think of it rather!  Will get back to this.

robrankin:
Okay, test written, let me know if there's any improvements to be had?

lib/facter/consul_version.rb
```diff
@@ -0,0 +1,15 @@
+# consul_version.rb
+
+Facter.add(:consul_version) do
+  setcode do
+    if Facter.value(:kernel) == 'Linux'
+      version = Facter::Util::Resolution.exec('consul --version')
+      version = version.lines.first.split[1].tr('v','')
+    elsif Facter.value(:kernel) == 'windows'
+      version = Facter::Util::Resolution.exec('"C:\\Program Files\\Consul\\consul.exe" --version')
```
split3: Is there anyway to make this configurable as we can't be certain that this is actually where consul is installed on Windows.  For instance we install to C:\Consul not Program Files.

robrankin: Agree, not optimal.  Since this module doesn't yet support Windows, maybe best to remove Windows support in the fact?  Only reason it's there currently is because I/we use Consul on Windows but use a separate module to install/manage it.

split3: If there is no easy way to parameterize it then yea might want to remove it.  My company is working on the Windows support for this module and are baking it in our preprod environments currently so hopefully we'll have that merged in soon.

robrankin: Don't think it would be possible to parameterize it, as facts run on the agent prior to catalogue compilation, so don't think they have access to any of the puppet parameters.  Would probably have to ensure consul was on the Windows path once support is added.

Also, I wrote the windows puppet module for my companies use of Consul on Windows, depends on NSSM.  Happy to contribute to the work you're doing if you're willing to put it in an open repo somewhere?

robrankin:
Removed Windows for now, will retain that on my own branch for our use.

spec/unit/facter/consul_version_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    context 'Returns consul version on Linux'
+    it do
+      consul_version_output = <<-EOS
+Consul v0.6.0
+Consul Protocol: 3 (Understands back to: 1)
+      EOS
+      allow(Facter.fact(:kernel)).to receive(:value).and_return("Linux")
+      allow(Facter::Util::Resolution).to receive(:exec).with('consul --version').
+        and_return(consul_version_output)
+      expect(Facter.fact(:consul_version).value).to match(/\d+\.\d+\.\d+/)
```
solarkennedy: This is fine, but can you add a really strong expect fact to eql '0.6.0' so I don't have to trust the regex logic? With such a controlled input, we should be able to say *exactly* what the output was, and not worry about spaces or anything extra.

lib/facter/consul_version.rb
```diff
@@ -0,0 +1,10 @@
+# consul_version.rb
+
+Facter.add(:consul_version) do
+  confine :kernel => 'Linux'
```
solarkennedy: ty for using confine

solarkennedy:
What is the expected output of this fact when consul isn't installed?
I would like the fact to not exist if consul is not installed. (behavior of most puppet facts)

spec/unit/facter/consul_version_spec.rb
solarkennedy: Bump on this one request, the only blocker to a merge.

robrankin: Been hectic, apologies for the delay.

solarkennedy:
Looks great! Thanks.

PR #212, opened 2016-01-05
Add support of OpenSuSE and SLED

kscherer:
All currently supported versions of OpenSuSE use systemd. Suse
Enterprise Linux has two flavors: Server and Desktop. The 11.x release
still uses sysv but the 12.0 release uses systemd.

Signed-off-by: Konrad Scherer <Konrad.Scherer@windriver.com>


solarkennedy:
Cool! Can you also amend `metadata.json`?

solarkennedy:
Thanks!

PR #213, opened 2016-01-06
Fix port property

horsehay:
Fix port property references in type provider and use defaultto syntax in type.


solarkennedy:
Look like this is covered in https://github.com/solarkennedy/puppet-consul/pull/214/ as well.
I'll go ahead and merge this in but @sigerber just fyi.

PR #214, opened 2016-01-06
ACL fixes - idempotentcy and port bug.

sigerber:
Implemented prefetch so that multiple puppet runs will not attempt to recreate the ACLs. Also fixed ACLs not working due to a typo introduced by the addition of the port parameter.

Note: I've only been playing with puppet for a couple of weeks so this isn't necessarily the best implementation. I'm also a bit uncomfortable about the reliance on 'port' and 'acl_api_token' needing to be correct across all tokens or bad things might happen.

### DIFF
--- a/lib/puppet/provider/consul_acl/default.rb
+++ b/lib/puppet/provider/consul_acl/default.rb
@@ -6,14 +6,35 @@ Puppet::Type.type(:consul_acl).provide(
 ) do
   mk_resource_methods
 
-  def self.list_resources(acl_api_token)
+  def self.prefetch(resources)
+    resources.each do |name, resource|
+      Puppet.debug("prefetching for #{name}")
+      port = resource[:port]
+      token = resource[:acl_api_token]
+
+      found_acls = list_resources(token, port).select do |acl|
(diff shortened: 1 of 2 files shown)


solarkennedy:
@sigerber does this actually make `puppet resource consul_acl` spit stuff out?

@michaeltchapman or @mdelagrange can you code review this?

sigerber:
Sorry, no. You need to implement 'self.instances' for that - but self.instances doesn't take any parameters, so there isn't a way AFAIK to feed in the port and ACL token. This just fixes the port and id bug and stops puppet trying to create the resource every time there is a puppet run.

mdelagrange:
seems reasonable to me

solarkennedy:
@sigerber can you merge master then I'll merge it in?

sigerber:
@solarkennedy Done - and hopefully did it right.

solarkennedy:
Thanks! I (personally) am not a user of this particular feature of this module, and we don't have a lot of test coverage on these, so we're going to depend on you and the other users on this thread to test this irl.

PR #219, opened 2016-01-20
Add hostname proprety for ACL operations, defaults to localhost, like before

### LINKED ISSUE #216: Support creating ACLs while talking to a hostname other than localhost
Right now, port is configurable, but the hostname is always localhost

gozer:
Fixes #216

### DIFF
--- a/lib/puppet/provider/consul_acl/default.rb
+++ b/lib/puppet/provider/consul_acl/default.rb
@@ -10,9 +10,10 @@ Puppet::Type.type(:consul_acl).provide(
     resources.each do |name, resource|
       Puppet.debug("prefetching for #{name}")
       port = resource[:port]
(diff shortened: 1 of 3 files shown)


split3:
If you could add tests that would be great for this one as well.  Also it would probably be fine to combine these two together into a single pull request if you'd like.  Otherwise we might end up with having to do some merge conflict resolution since this and your other are changing several of the same lines of code.

gozer:
Tests will be on their way soon. As for the merge conflicts, I'll handle them, I'd rather keep logically different changes in different PRs

solarkennedy:
Looks good to me. @split3 second opinion?

split3:
Yes agree, looks good, thanks for adding the test!

solarkennedy:
Thanks!

PR #222, opened 2016-01-28
Debian init script should depend on networking, resolution and syslog

chrisboulton:
I've had these changes sitting out in a branch for a while and have been meaning to submit them back.

I noticed we had issues with Consul starting in the wrong order on our systems - primarily because we were using syslog for logging and the syslog daemon hadn't been started by the time Debian had decided to start Consul.


solarkennedy:
Thanks!

PR #227, opened 2016-02-08
lazily return nil when consul client isn't installed

roobert:
This patch fixes the following error which occurs when consul isn't installed on a client:

```
Error: Facter: error while resolving custom fact "consul_version": undefined method `lines' for nil:NilClass
```


solarkennedy:
Thanks!

PR #229, opened 2016-02-09
Fix fact when consul isn't installed

jaxxstorm:
If consul isn't installed, I get the a nil:NilClass error:

```
Could not retrieve fact='consul_version', resolution='<anonymous>': undefined method `lines' for nil:NilClass
```

This fixes that.


jaxxstorm:
Someone fixed this before me, and in a better way!

PR #233, opened 2016-02-26
Added custom init style "custom"

ph0enix49:
This is useful for scenarios when you want to start consul from another program manager, e.g supervisor (supervisord.org).

Full example:
  class { 'consul':
    package_ensure => $package_ensure,
    config_hash        => $config_hash,
    init_style             => 'custom',
  }

  supervisor::program { 'consul':
    command   => '/usr/local/bin/consul agent -config-dir /usr/local/consul/conf',
    user            => 'consul',
    group          => 'consul',
  }

This change will prevent puppet-consul from creating an init script and managing the process, so the supervisor could do this instead.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -78,9 +78,6 @@ class consul::config(
           content => template('consul/consul.launchd.erb')
         }
       }
-      'custom' : {
-        notice("I'm not creating an init script for you as you are going to create one yourself")
-      }
       default : {
         fail("I don't know how to create an init script for style ${consul::init_style}")
       }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -78,6 +78,9 @@ class consul::config(
           content => template('consul/consul.launchd.erb')
         }
       }
+      'custom' : {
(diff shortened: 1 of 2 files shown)


manifests/config.pp
```diff
@@ -78,6 +78,9 @@
           content => template('consul/consul.launchd.erb')
         }
       }
+      'custom' : {
+        notice("I'm not creating an init script for you as you are going to create one yourself")
```
solarkennedy: No notice please. This causes puppet to never converge.

solarkennedy:
Can't people already do this if they say `initstyle => false`  and `manage_service => false` ?

ph0enix49:
- If `manage_service => false` then `reload_service` won't work, i.e. I still want consul to reload itself whenever there's a config change.

- `init_style => false` is actually works, and there's no need for "custom". I reduced the diff and now th is what we need.

solarkennedy:
Gotcha.

ph0enix49:
Thanks!

PR #234, opened 2016-02-28
add support for arm (Raspberry pi's ARM here) architecture

gibre:
Available since 0.6.0 on the mirror used in the module (https://releases.hashicorp.com/consul/)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -20,7 +20,7 @@ class consul::params {
   case $::architecture {
     'x86_64', 'amd64': { $arch = 'amd64' }
     'i386':            { $arch = '386'   }
-    'armv61':          { $arch = 'arm'   }
+    /^arm.*/:          { $arch = 'arm'   }
     default:           {
       fail("Unsupported kernel architecture: ${::architecture}")
     }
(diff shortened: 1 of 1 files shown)


solarkennedy:
Cool. Do you know if you can regex it and do `^arm`? I don't want to get into the business of enumerating all the arm arch's and I be they cross compile the consul binary for many of them.

gibre:
Yeah good idea, indeed, enumerating all arm architectures is not really efficient :)

EvanKrall:
arm64, armel, armhf are not necessarily compatible.

solarkennedy:
@EvanKrall Hmmm. Here is a file on the binary:

    consul: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, not stripped

I don't know how nuts we have to be about this. I'll take a second opinion, should we just glob arm and leave it up to the user to know their own hardware? Or is there a sane regex we can do to capture the arches we want but not the other ones? (my vote is for the glob)

split3:
I'd say glob it, let a future pull request from somebody that might provide a better regex improve it if needed.

solarkennedy:
Globbed.

Issue #231, opened 2016-02-22
Service reload too quick


nvtkaszpir:
Looks like sometimes Exec['reload consul service'] executes commands while consul service is still starting up - thus getting error message 'connection refused'.

For me  this was fixed by adding 'tries' to Exec in reload_service.pp

```puppet
    exec { 'reload consul service':
      path        => [$consul::bin_dir,'/bin','/usr/bin'],
      command     => "consul reload -rpc-addr=${consul::rpc_addr}:${consul::rpc_port}",
      tries       => 3,
      refreshonly => true,
    }
```

But still needs better testing.

PR #235, opened 2016-03-04
Update reload_service.pp

nvtkaszpir:
Solves reload failures if consul service is not fully operational.

https://github.com/solarkennedy/puppet-consul/issues/231


solarkennedy:
Hmm. Seems like an ok solution to this. Thanks!

PR #238, opened 2016-03-11
Fix a puppet 2.7 compatibility issue

joshuaspence:
It seems that `-> file` doesn't work if multiple `file` resources are specified within the same block. On puppet 2.7.10, I get the following error:

```
Could not find resource 'File[/opt/staging/consul-0.6.3/consul]File[/usr/local/bin/consul]' for relationship from 'Staging::Extract[consul-0.6.3.zip]'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,11 +12,11 @@ end
 
 group :test do
   gem "json"
-  gem "rake"
   gem "puppet", ENV['PUPPET_VERSION'] || '~> 3.7.0'
   gem "puppet-lint"
 
   # Pin for 1.8.7 compatibility for now
+  gem "rake", "< 11.0.0"
   gem "rspec", '< 3.2.0'
   gem "rspec-core", "3.1.7"
   gem "rspec-puppet", "~> 2.1"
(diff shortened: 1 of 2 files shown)


joshuaspence:
The Travis failures don't seem related.

joshuaspence:
Here is a MWE:

```
> puppet apply -e "notify { 'foo': } -> notify { 'bar': message => 'bar'; 'baz': message => 'baz'; }"
Could not find resource 'Notify[bar]Notify[baz]' for relationship from 'Notify[foo]'
```

solarkennedy:
Well... they are not directly related to the change, but it does look like newer versions of rake have dropped support for ruby 1.8.7.

I initially resisted support for these older versions of ruby/puppet. I'm happy to merge but you will have to adjust the gemfile for backwards compatibility. (pin Rake?)

split3:
I don't believe we need this anymore as I think the support was already added, and I pinned the gemfile a little while ago for those tests on 2.7 to pass.  If this is still needed please rebase and re-open and we can take a look.

PR #241, opened 2016-03-15
EL < 7 uses init service provider, not sysv. 

### LINKED ISSUE #240: Master broken on EL6 due to "Invalid service provider 'sysv'"
When mistakenly using master, I observed that this module was broken on my CentOS 6 test box with the following error:
> Invalid service provider 'sysv' at /modules/consul/manifests/run_service.pp:19

A quick browse through the code suggests that 85de0cde707f2c011754677c097c53e61e66e664 is the cause of this breakage, specifically whereby it's enforcing the `provider` param.
On <EL7, the default `provider` is `init`, not `sysv`. 

Running through the history on `params.pp`, it looks like the this was first entered into `params.pp` in 24931c85f9257d42f5f921d047a1b51ec43cf03e, however as it was never passed through the service, it was never an issue previously... 

PR incoming to correct default value on <EL7... 

Cheers
Gav

fatmcgav:
Updates to use init service provider on EL < 7, renamed any sysv reference to init. 

Fixes #240


solarkennedy:
Thanks!

PR #243, opened 2016-03-17
Switching from staging to archive module

split3:
Since the staging module won't be receiving any TLC anymore we'd probably do well to upgrade to the module that replaces it, the puppet/archive module.  This will provide checksum support down the road (currently somewhat experimental with urls).


solarkennedy:
This looks great! I'm happy we can finally move to the `archive` module.

@aj-jester had previously attempted this and found some gotchas:
https://github.com/solarkennedy/puppet-consul/pull/178

It sounds like this may be potentially disruptive to some, but I think long term we have to do it.

split3:
The only gotcha I think we need to be concerned with is the _first run_ one.  And I think we can get around this by specifying the install path to be the same one that is used previously by staging **/opt/staging**.  This is specified by the _install_path_ variable that is set.  So I think that would probably get us around that issue if that is the only major concern.

It would be possible for us to also possibly build in a check for the staging folder path and if that version exists let everything download and relink but just not restart.  That should be adequate as well.

I'm open to either solutions you guys want to go with.  It might be more challenging changing the notify as it might not be in quote and accessible, but we could see about cancelling the refresh (not sure I've done that before though)

split3:
Another option would be to use the consul_version fact we have to see if it is the same version that is being downloaded and relinked and not reload if it is.

split3:
Ok added additional tests and rebased.

split3:
Also tested this on a few systems and archive downloaded the artifact and extracted, and updated the symlink but didn't restart the service.  So I think we've covered the biggest concern.

solarkennedy:
@split3 I've uploaded a new commit to this branch that fixes the integration tests. (`bundle exec rake beaker)

It looks like thew web_ui download gets installed into `/opt/puppet-archive/consul-0.6.4_web_ui` can you confirm this is correct? If so I say we ship.

split3:
@solarkennedy Yes I kept the same functionality we had previously to use the symlinks for both the executable along with the ui so that is expected.

split3:
@solarkennedy Thanks for fixing the acceptance tests, any reason we don't run those are part of travis builds?  Sorry wasn't running those locally since I didn't see them as part of travis.

split3:
@solarkennedy so are we good here now?

solarkennedy:
I'm very sick right now and I don't have a recovery eta. Please continue
and use standard best judgment. Thank you for helping in my absence.
Zzzzzz....

split3:
Sorry to hear @solarkennedy, rest up and get better.  I'll go ahead and merge this in and we can make adjustments as needed.

PR #244, opened 2016-03-18
Changing systemd script due missing environment

Karen09:
hi
I have decided that systemd template has missed environment option , about user and group section we can discuss why to put there user and group credentials ?

Regards 
Karen

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,6 +34,12 @@ class consul::config(
         }
       }
       'systemd' : {
+        file { '/etc/sysconfig/consul':
+          ensure  => directory,
+          mode    => '0644',
+          owner   => 'root',
+          group   => 'root',
+        }->
         file { '/lib/systemd/system/consul.service':
           mode    => '0644',
           owner   => 'root',
--- a/templates/consul.systemd.erb
+++ b/templates/consul.systemd.erb
@@ -1,16 +1,14 @@
 [Unit]
-Description=Consul Agent
-Wants=basic.target
-After=basic.target network.target
+Description=Consul agent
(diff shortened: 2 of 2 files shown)


manifests/config.pp
```diff
@@ -34,6 +34,12 @@
         }
       }
       'systemd' : {
+        file { '/etc/sysconfig/consul':
+          ensure  => directory,
```
solarkennedy: I don't understand, is this intended to be a file or a directory?

Karen09: Hi 
Yes exactly its need to be file , sorry it is my fail.
Regards
Karen

templates/consul.systemd.erb
```diff
@@ -1,16 +1,14 @@
...
 
 [Service]
-User=<%= scope.lookupvar('consul::user') %>
-Group=<%= scope.lookupvar('consul::group') %>
-ExecStart=<%= scope.lookupvar('consul::bin_dir') %>/consul agent \
-  -config-dir <%= scope.lookupvar('consul::config_dir') %> <%= scope.lookupvar('consul::extra_options') %>
-ExecReload=/bin/kill -HUP $MAINPID
-KillMode=process
+EnvironmentFile=-/etc/sysconfig/consul
 Restart=on-failure
```
Karen09: Hi
 One more thing that we have to look after consul installation if in system "systemctl daemon-reload " command is used  then i do not get consul.service  automatically restart ,so service consul is hanging.thanks for collaboration

Regards
Karen

solarkennedy:
Can you merge master and fix the test/lint failures reported by travis?

Issue #236, opened 2016-03-08
Error while trying to create user consul... due group already exists


brahama:
Could not create user consul: Execution of '/usr/sbin/useradd -M -r consul' returned 9: useradd: group consul exists - if you want to add this user to that group, use -g

Just tried and removed the group to see if that perhaps already existed, but same issue as group is created first.

brahama:
Yes, it can be bypassed by the manage_group to false. But seems to me its not that tidy perhaps.

Cheers!

solarkennedy:
Ah. Seems like we should be ordering the group before the user in the install.pp. Can you make a PR?

Cinderhaze:
From what I understood, users should autorequire groups if they are declared in the same manifest.. I may be mistake though

solarkennedy:
Yea, but only primary group (gid) and that is what #247 solves.

If you care about this issue, would you like to rebase #247 so it passes tests and applies on master?

PR #247, opened 2016-03-24
new condition and gid reference added to user

patant:
#236 Problem is that when adding a user and a group the user needs to be added with the gid to the group.

Solution added a new condition when manage_user and manage_group is true. The group created is added to the user during the creation of user. If the condition is not met the resources will be created same way as before.

Tested on Debian and Centos.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -99,21 +99,31 @@ class consul::install {
     }
   }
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
Can you merge/rebase on master and address test failures?

PR #248, opened 2016-03-29
Use redhat init instead of "init" init

benpillet:
https://docs.puppetlabs.com/puppet/3.8/reference/types/service.html#service-provider-redhat

I had CentOS 6 nodes in an autoscaling group.  When they would get scaled down/terminated, consul would show them as failed.  I had leave_on_terminate set to true, but I was still seeing the error.  When I ran chkconfig on the box, the consul service wasn't listed.  So, the K20consul script was not getting called in /etc/rc6.d and the node wasn't leaving the cluster.  Please correct me if I'm wrong, but I suspect this provider is the right one for the operatingsystem entries listed.  Also, is this even necessary as puppet is supposed to pick the "best" init_style for a service anyway.  Should most of these elsif entries just be removed?

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,7 +38,7 @@ class consul::params {
     }
   } elsif $::operatingsystem =~ /Scientific|CentOS|RedHat|OracleLinux/ {
     if versioncmp($::operatingsystemrelease, '7.0') < 0 {
-      $init_style = 'init'
+      $init_style = 'redhat'
     } else {
       $init_style  = 'systemd'
     }


benpillet:
I should've actually tried this patch on my box.  Looks like it doesn't work because of config.pp line 82.  I commented out the provider line in run_service.pp.  Is the provider necessary?  i.e. Commenting it out on CentOS 6 works, but isn't clean for all the other platforms.  Any reason not to trust the default puppet service provider?

solarkennedy:
Yea, we basically can't make everyone happy. See https://github.com/solarkennedy/puppet-consul/pull/125

cliles:
@solarkennedy - #125 also introduces a break for Amazon linux.

cliles:
init provider is a poor choice, it does not support enable.

https://docs.puppet.com/puppet/latest/reference/type.html#service-providers

Perhaps #125 should have only introduced changes for debian via an if block?

solarkennedy:
Well, if we know we are on redhat then I guess it should be safe to use the redhat provider (baring systemd)

I'm convinced. @benpillet I'll merge this in as soon as you fix the tests and rebase.

gasparch:
fixed problems for me as well on CentOS 6.8

PR #254, opened 2016-04-20
in dynamic environment consul-agent should be init

mcortinas:
I modified this value in my puppet repo after check consul-agent was finished with clean exit or somethig of these signals SIGHUP, SIGINT, SIGTERM or SIGPIPE

I need consul is always working in live "nodes"

I also read this documentation: https://www.freedesktop.org/software/systemd/man/systemd.service.html


split3:
I think it might make better sense that this be configurable for systemd.

solarkennedy:
I think this is totally fine, we are already doing ensure=>running in most cases, so puppet is going to be doing this anyway just at a slower pace.

PR #255, opened 2016-04-26
Puppet archive install directory will fail if default umask changed

lynxman:
This fix will ensure the directory is always available to the consul user, I've observed this issue when securing a server and setting a default umask 077 rather than the more permissive umask 022.


solarkennedy:
Thanks

PR #257, opened 2016-05-26
adding ability to define the inital status of consul checks

asgolding:
Would like to add the ability to define the inital check status as described in https://www.consul.io/docs/agent/checks.html under the heading "Initial Health Check Status".

I've tested this change and it works.


asgolding:
I've had a look at the test errors and they don't appear to be related to my change.

solarkennedy:
Can you add docstring?

solarkennedy:
ty

PR #260, opened 2016-06-01
Added a parameter to allow users to change $install_path

tfhartmann:
In the off chance /opt is not available for use, this PR allows you to set archive_path, which is in turn used in consul::install when installing via the URL.


solarkennedy:
Looks great, ty.

PR #262, opened 2016-06-02
drop bool2str as we need a unquoted true|false

### LINKED ISSUE #261: enableTagOverride value being written as a quoted string, need to be unquote boolean.
The recent addition of the enableTagOverride service config option seems to be writing the value as a quoted string:

```
"enableTagOverride":"false"
```

Which cause consul to fail to load the service:

```
Jun  2 11:06:09 dev01 consul: ==> Error decoding '/etc/consul/service_slurmctld.json': 1 error(s) decoding:                                                               │·
Jun  2 11:06:09 dev01 consul: * 'EnableTagOverride' expected type 'bool', got unconvertible type 'string'
```

Removing the quotes around false, allows it to load: 

```
"enableTagOverride":false
```

I'll see if i can work up a fix and submit a pr, but wanted to file the issue now.

sjoeboo:
fixes #261 

does a little puppet-lint clean up as well.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -44,7 +44,7 @@ define consul::service(
   $token               = undef,
   $enable_tag_override = false,
 ) {
-  include consul
+  include ::consul
 
   consul_validate_checks($checks)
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
ty

PR #263, opened 2016-06-04
Handle nested :undef in consul_sorted_json

mcasper:
Currently `consul_sorted_json` only handles top level `:undef` values,
leaving nested `:undef` values to slip through and eventually raise an
exception. Instead of trying to handle all `:undef`s upfront in the
function, handle them in the recursion and treat them as what they
actually are, `null`s.

I've also been encountering the `Exception("")` vs `Exception.new("")`
problem detailed in #245, as `Exception("")` is not valid syntax in
newer versions of Ruby. That fix is included here as well


solarkennedy:
Nice. Thanks!

PR #264, opened 2016-06-08
Update consul_validate_checks.rb

Lswith:
As per the docs (https://www.consul.io/docs/agent/checks.html) a TTL check looks like this:
```
{
  "check": {
    "id": "web-app",
    "name": "Web App Status",
    "notes": "Web app does a curl internally every 10 seconds",
    "ttl": "30s"
  }
}
```


Lswith:
this was unnecessary.

PR #265, opened 2016-06-14
Configure log file for upstart.

jdavisp3:
This PR configures a log file for the upstart configuration similar to the one for the [`init.d`](https://github.com/solarkennedy/puppet-consul/blob/master/templates/consul.init.erb#L19) configuration.

I tested this by provisioning a consul server with the new configuration and observing the log file being created log entries being appended. I restarted the service and the new entries were appended to the existing file.

### DIFF
--- a/templates/consul.upstart.erb
+++ b/templates/consul.upstart.erb
@@ -10,6 +10,7 @@ env GROUP=<%= scope.lookupvar('consul::group') %>
 env DEFAULTS=/etc/default/consul
 env RUNDIR=/var/run/consul
 env PID_FILE=/var/run/consul/consul.pid
+env LOG_FILE=/var/log/consul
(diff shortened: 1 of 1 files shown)


solarkennedy:
But if you are running consul as non-root, will it be able to continue to log? I guess it doesn't really log to a file anyway.

jdavisp3:
Thanks for taking my contribution!

> But if you are running consul as non-root, will it be able to continue to log? I guess it doesn't really log to a file anyway.

In my test setup the `consul` Java process does not run as root but it is able to log to the file. The upstart configuration will create the log file as root and set up the I/O redirection before launching the `start-stop-daemon` process which switches to the `consul` user.

Note this configuration does log to a file: `/var/log/consul`.

PR #266, opened 2016-06-17
Fix windows_service so it works correctly

lynxman:
Just fixed a couple escapes needed for this to work

### DIFF
--- a/manifests/windows_service.pp
+++ b/manifests/windows_service.pp
@@ -22,8 +22,9 @@ class consul::windows_service(
 
   $app_dir = regsubst($consul::bin_dir, '\/', '\\', 'G')
   $app_exec = "${app_dir}\\consul.exe"
-  $agent_args = regsubst($consul::config_dir, '\/', '\\', 'G')
-  $app_args = "agent -config-dir=${agent_args}"
+  $configdir_args = regsubst($consul::config_dir, '\/', '\\', 'G')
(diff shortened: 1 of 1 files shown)


manifests/windows_service.pp
```diff
@@ -52,7 +53,7 @@
   } ->
   exec { 'consul_service_set_parameters':
     cwd         => $consul::bin_dir,
-    command     => "${consul::bin_dir}/set_service_parameters.ps1",
+    command     => "& \"${consul::bin_dir}/set_service_parameters.ps1\"",
```
solarkennedy: Can you describe the "`& `" prepend thing?

lynxman: The ampersand (the 'and' sign) here tells PowerShell to execute that command, instead of treating it as a cmdlet or a string. The backticks (the funny looking single-quotes) are there to escape the following character, similar to the \" in C-based languages, or double-double-quotes ("") in VB. Otherwise the " character will end the string and the parser will cry when it can't understand what you're trying to say after that. (You can alternatively use single-quotes instead in this case, as I have in the previous example.)
http://edgylogic.com/blog/powershell-and-external-commands-done-right/
Just want to ensure this is executed as a command inside powershell, this is the way to force it to do so

solarkennedy:
Ok. We don't currently have any unit or itests that are windows-specific, so we'll have to depend on you and the other windows users to test this and sure it says fixed. Thank you for the contribution!

PR #146, opened 2015-06-30
upstart: Agents should gracefully leave cluster on stop

jsok:
Similar to PR #87  but for upstart.

Tested with Ubuntu 14.04:

```
Jun 30 12:30:49 ip-10-0-3-159 puppet-user[19088]: (Class[Consul::Run_service]) Scheduling refresh of Service[consul]
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: agent.rpc: Accepted client: 127.0.0.1:32928
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: agent.rpc: Accepted client: 127.0.0.1:32929
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: agent.rpc: Graceful leave triggered
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: consul: client starting leave
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: serf: EventMemberLeave: ip-10-0-3-159 10.0.3.159
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: agent: requesting shutdown
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: consul: shutting down client
Jun 30 12:30:49 ip-10-0-3-159 consul[18056]: agent: shutdown complete
Jun 30 12:30:49 ip-10-0-3-159 puppet-user[19088]: (/Stage[main]/Consul::Run_service/Service[consul]) Triggered 'refresh' from 1 events
Jun 30 12:30:49 ip-10-0-3-159 consul[19801]: serf: EventMemberJoin: ip-10-0-3-159 10.0.3.159
Jun 30 12:30:49 ip-10-0-3-159 consul[19801]: agent: failed to sync remote state: No known Consul servers
Jun 30 12:30:49 ip-10-0-3-159 consul[19801]: agent: Joining cluster...
Jun 30 12:30:49 ip-10-0-3-159 consul[19801]: agent: (LAN) joining: [consul.staging.local]
```

PR #267, opened 2016-06-21
Added KillSignal to gracefully halt agent

idarnius:
In order to use, in systemd service, kill procedure configuration _KillMode=process_ (default signal sent is _SIGTERM_ https://www.freedesktop.org/software/systemd/man/systemd.kill.html# ) and be able to stop consul agent gracefully. We would have to use configuration option _KillSignal_ sent _SIGINIT_ (https://www.consul.io/docs/agent/basics.html)

### DIFF
--- a/templates/consul.systemd.erb
+++ b/templates/consul.systemd.erb
@@ -9,7 +9,9 @@ Group=<%= scope.lookupvar('consul::group') %>
 ExecStart=<%= scope.lookupvar('consul::bin_dir') %>/consul agent \
   -config-dir <%= scope.lookupvar('consul::config_dir') %> <%= scope.lookupvar('consul::extra_options') %>
 ExecReload=/bin/kill -HUP $MAINPID
+<% if @config_hash["server"] == false -%>
 KillSignal=INT
+<% end -%>
 KillMode=process
 Restart=always
 RestartSec=42s
--- a/templates/consul.systemd.erb
+++ b/templates/consul.systemd.erb
@@ -9,6 +9,7 @@ Group=<%= scope.lookupvar('consul::group') %>
 ExecStart=<%= scope.lookupvar('consul::bin_dir') %>/consul agent \
   -config-dir <%= scope.lookupvar('consul::config_dir') %> <%= scope.lookupvar('consul::extra_options') %>
 ExecReload=/bin/kill -HUP $MAINPID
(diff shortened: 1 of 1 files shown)


solarkennedy:
Hmm, I'm not 100% sure this is desired behavior. Check out #146 and #87.

The general consensus was that it was dangerous for non-agents to leave a cluster gracefully.
Can you (is it possible in systemd) make this PR match the other init script behavior? (only gracefully leave cluster if agent)

idarnius:
I have added conditional in the template to ensure that stop consul agent gracefully only apply for consul agent client and no consul agent server.

However Consul Agent Basics ( https://www.consul.io/docs/agent/basics.html ) said:

> It is especially important that a server node be allowed to leave gracefully so that there will be a minimal impact on availability as the server leaves the consensus quorum.

> For client agents, the difference between a node failing and a node leaving may not be important for your use case. For example, for a web server and load balancer setup, both result in the same outcome: the web node is removed from the load balancer pool.

solarkennedy:
Hmm. These lint failures in travis are not from this PR, I don't know what is up with that. Don't worry about them.

Also, per the docs, are we "doing it wrong"? (wrt agent versus server leaving gracefully)

solarkennedy:
Also please merge/rebase on master to fix the tests.

idarnius:
From my point of view when you stop the service (in this case with systemd) the expected behaviour would be the node ( agent with server role or client role) leave the cluster (as doc said with a signal SIGINT).

Docs emphasizes in that is _especially important that a server node be allowed to leave gracefully_ So I think that always we use SIGINT to stop consul agent (server or not). Other signals in some cases could produces _false_ failing states. For me it is important that I can control the state of services, and I need to deregistrate services when I stop consul agent. 

> When a node leaves, it specifies its intent to do so, and the cluster marks that node as having left. Unlike the failed case, all of the services provided by a node are immediately deregistered. If the agent was a server, replication to it will stop.

Through the proper config you can rejoin consul agent (server o client) to cluster when the service starts again the service.

solarkennedy:
Just want to let you know I haven't forgotten about this PR.

solarkennedy:
This behavior has changed since the opening of this PR. The default now is to *not* leave the cluster regardless of server or client.

Issue #275, opened 2016-08-25
Support for puppet-archive >=1.0.0


mgla:
The current version of puppet-archive is v1.1.1, and it _seems_ to work just fine with v.1.1.1 (I am not able to verify this). Maybe update the metadata.json to allow newer archive versions without error?

Current dependency:
```
    {
      "name": "puppet/archive",
      "version_requirement": ">=0.5.0 <1.0.0"
    }
```

epleterte:
I hit this now as well. From puppet-librarian:

    [Librarian] Conflict between puppet-archive (< 1.0.0, >= 0.5.0) <(no source specified)> and puppet-archive/1.0.0 <https://forgeapi.puppetlabs.com>
    Could not resolve the dependencies.

I will try just allowing a newer version of the archive module.

ghoneycutt:
Tried using puppet/archive v1.1.1 and it has syntax errors that made it unusable at all with Puppet v3.

mgla:
@ghoneycutt Can you elaborate on the syntax errors? It works fine for me

solarkennedy:
The only build failures in #227 are related to Puppet 3.1.0.
I'm pretty sure we have test coverage to cover all the cases where we use `archive`.

krjensen:
Can we get this fixed soon? unable to install on Ubuntu 16.04 without this fix.

solarkennedy:
@ghoneycutt can you look at #277? I did have to remove 3.1.0 from the matrix, the exact failure was here: https://travis-ci.org/solarkennedy/puppet-consul/jobs/155769754

The matrix on that pr still has 3.3, 3.7, and 3.8, so decent v3 coverage?

solarkennedy:
I've release 1.1.0 of this module with this change.

PR #276, opened 2016-08-28
Allow for stable versions of puppet/archive

ghoneycutt:
Without this patch, the current metadata instructs that only unstable versions before 1.0.0 are supported. Since those versions allow for breaking changes, we should allow for the stable v1.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -15,7 +15,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">=0.5.0 <1.0.0"
+      "version_requirement": ">=0.5.0 <2.0.0"
     }
   ],
   "operatingsystem_support": [


ghoneycutt:
This addresses issue #275

solarkennedy:
Got it. I'll bump fixtures so we start testing against new versions.

ghoneycutt:
Do not merge yet as puppet/archive v1 does not work with this module.

ghoneycutt:
just beat me to it

ghoneycutt:
Suggest digging in on the changes to the module and just supporting v1

solarkennedy:
Ok. I won't make a new release yet until we are sure, but at least you can install it easily to test.

If you do know what doesn't work, add some notes to #275 or PR or whatever.

ghoneycutt:
Thanks!

PR #278, opened 2016-09-15
Polish

tdevelioglu:
A little polish. Complete parameter docstring, move defaults to params and scope variables.


solarkennedy:
Thanks!

PR #279, opened 2016-09-15
Decouple service from init system

tdevelioglu:
Decouple managing service and init system from eachother.

Previously the logic didn't make sense, where one could pass a bool
to the parameter `init_style` that expects a string to toggle the
management of service files in the init system and it was impossible to
let the module manage the service but not the init system.

This breaks the common case of where you want to let a package prep the
init system but still let the module manage the service.

This commit decouples the service resource from management of the init
system so it offers the choice of using the module-supplied init files.
It adds support for a magic value `unmanaged` on parameter `init_style` that
disables management of the init system.


solarkennedy:
This PR includes commits from the "polish" one, can you rebase?

tdevelioglu:
Done

solarkennedy:
I understand what you are trying achieve.

What if instead of a new parameter we reused existing the `init_style`, and if users don't want the init stuff managed they can say `init_style => 'unmanaged'` ? (and of course configure the case statement to match, `provider` would be left to undef to let puppet autodetect what the package provided.)

tdevelioglu:
Sounds good to me, how is this ?

solarkennedy:
Thanks!

PR #285, opened 2016-10-04
Remove puppet error when ACLs cannot be retrieved

thejandroman:
Currently if consul does not come up or there is some issue electing a
leader, the consul_acl will throw a hard puppet error. This causes the
puppet run to immediately fail. This commit changes that puppet error to
a warning. The unavailability of the consul API should not cause a
puppet failure.

### DIFF
--- a/lib/puppet/provider/consul_acl/default.rb
+++ b/lib/puppet/provider/consul_acl/default.rb
@@ -57,7 +57,9 @@ Puppet::Type.type(:consul_acl).provide(
     if res.code == '200'
       acls = JSON.parse(res.body)
     else
-      raise(Puppet::Error,"Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
+      Puppet.warning("Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/consul_acl/default.rb
```diff
@@ -57,7 +57,9 @@ def self.list_resources(acl_api_token, port, hostname, protocol, tries)
     if res.code == '200'
       acls = JSON.parse(res.body)
     else
-      raise(Puppet::Error,"Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
+      Puppet.warning("Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}")
+      $stderr.puts "Cannot retrieve ACLs: invalid return code #{res.code} uri: #{path} body: #{req.body}"
```
solarkennedy: Do you need both the puppet warning and stderr?

thejandroman: No. That must have snuck in during testing. I'll remove the stderr.

solarkennedy:
I don't know whats up with those test failure, let me look at those a bit and compare to master.

PR #292, opened 2016-10-26
BREAKING: Change the default 'archive_path' to '/opt/consul/archives'.

jmkeyes:
If another Puppet module already defines the archive directory then compiling a catalog leads to a duplicate resource error; this patch should fix that issue.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,7 +5,7 @@
 #
 class consul::params {
   $acls                  = {}
-  $archive_path          = '/opt/puppet-archive'
+  $archive_path          = '/opt/consul/archives'
   $bin_dir               = '/usr/local/bin'
   $checks                = {}
   $config_defaults       = {}
--- a/spec/acceptance/standard_spec.rb
+++ b/spec/acceptance/standard_spec.rb
@@ -35,7 +35,7 @@ describe 'consul class' do
     end
 
     describe file('/opt/consul/ui') do
-      it { should be_linked_to '/opt/puppet-archive/consul-0.6.4_web_ui' }
+      it { should be_linked_to '/opt/consul/archives/consul-0.6.4_web_ui' }
     end
 
     describe service('consul') do
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -86,9 +86,9 @@ describe 'consul' do
   end
 
   context "When installing via URL by default" do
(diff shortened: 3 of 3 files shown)


solarkennedy:
This is the `archive_path`, I wouldn't expect people to be managing this outside of this module. Can you tell me more about your use-case and why this path is managed in your environment?

jmkeyes:
This module defines the `$archive_path` as `/opt/puppet-archive` by default and manages that directory. That path can also be managed by other modules that use [`puppet/archive`](https://github.com/voxpupuli/puppet-archive) and doesn't seem to be consul-specific. This patch modifies the module to avoid any duplicate resource declarations.

solarkennedy:
Hmm. You are right, that isn't really a good path for this module to be using.
But I don't think !defines are the way to go.

How about changing the default path to be `/opt/consul/` and this module can assume it totally control this dir and doesn't have to share it?

jmkeyes:
I've replaced the commit I made with one that updates `archive_path` to default to `/opt/consul` as requested. Will that work?

solarkennedy:
Sure, but *why* do you want to keep downloaded packages in a subfolder of /opt/consul?

jmkeyes:
The primary reason is for maintenance: I want to be able to automatically remove old archives that are no longer used after an installation or upgrade. Currently, the module stores the downloaded archives beside the folders that it extracts them into, which makes slightly harder to see what archives are unused.

solarkennedy:
sgtm, if we are going to change it, let's change it to something that can be purged automatically. Go ahead and set it to /opt/consul/archive(s) ?

solarkennedy:
@jmkeyes although now we have a new problem: it is unlikely that users will have /opt/consul ready for puppet to make the archives folder in :( (right?)

jmkeyes:
That is correct.

The examples within the README for this module specify `/opt/consul` as reasonable default for `$::consul::config_hash['data_dir']`. This hash key is then used in the `consul::install` class as a base directory for the extracted archive contents.

I'll update my patch to store archives under a subdirectory of `$::consul::data_dir` instead.

solarkennedy:
Ah, `data_dir`, right. Yes please. I think this is probably the most direct way to do it, as we *know* this folder must exist (eventually). 90% of the time that will be `/opt/puppet` anyway in the base, and then `archives` under that sounds very sane.

jamtur01:
Looks like this branch was never updated to reflect this: `Should I update my patch to store archives under a subdirectory of $::consul::data_dir instead?`. See #307.

Issue #291, opened 2016-10-25
New prepared queries provider not accepting integers


jaxxstorm:
My code
```
class roles::puppetmaster(
  $consul_failover_n = 1
) {
  # this works, so the integer is valid
  validate_integer($consul_failover_n)

consul_prepared_query { 'puppetmaster':
    ensure                => 'present',
    service_name          => 'puppetmaster',
    service_failover_n    => $consul_failover_n,
    service_failover_dcs  => $consul_failover_dcs,
    service_only_passing  => true,
    service_tags          => ['puppet'],
    ttl                   => 10,
  }
}
```

The result of this is:

```
# puppet agent -t --environment=consul_prepared_query --server=tukd-puppet-1.dapt.to
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for <hostname>
Error: Failed to apply catalog: Parameter service_failover_n failed on Consul_prepared_query[puppetmaster]: Nearest failover datacenters must be an integer at /etc/puppet/environments/consul_prepared_query/site/roles/manifests/puppetmaster.pp:158
```

Has anyone actually managed to get this working?

solarkennedy:
This feature is brand new. You might need to collaborate with @djtaylor to get things going.

You could do the `=> 0 + 10` hack to coerce it into an int maybe?

djtaylor:
I will try to take some time to look at it this week. I am storing integers in hieradata for that parameter, and passing it off to the provider via create_resources without that issue, but I will try it your way and see if I can work out a solution.

djtaylor:
@jaxxstorm I am unable to reproduce your error. I have tried passing in values to `service_failover_n` via hieradata, profile parameters, variables inside the class, and direct integers. Can you let me know what version of Puppet you are on?

solarkennedy:
Also try:
`service_failover_n    => $consul_failover_n + 0,`

jaxxstorm:
@djtaylor I'm on 3.8.7

asmajlovic:
@djtaylor Late to the show, I know, but I have seen this as well and figured I'd save others some hassle.  The issue is with `puppet-stdlib`, specifically the `is_a` function.  The function in question is fully supported in Puppet >= 4 and Puppet 3 only when `parser=future` is set for the environment (as per the relevant README):

### **is_a**

```
Boolean check to determine whether a variable is of a given data type.
This is equivalent to the =~ type checks.
[...]

PR #294, opened 2016-11-04
KV Provider / Prepared Query Bugfixes

djtaylor:
This adds a new KV provider for accessing the Key/Value database API. This also attempts to address issues in the previous Prepared Query PR and the following issues:
 
  - https://github.com/solarkennedy/puppet-consul/issues/291

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,18 @@ consul_prepared_query { 'consul':
 
 This provider currently only has support for basic prepared queries (not templated queries).
 
+## Key/Value Objects
+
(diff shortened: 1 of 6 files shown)


djtaylor:
Initial testing seems to confirm that the following works:

  - Create a new KV object
  - Update an existing KV object
  - Delete a KV object
  - Handle the 404 when no KV objects exist for the DC

solarkennedy:
Whoa! Thanks!

PR #295, opened 2016-11-07
Use `hiera_hash` with `create_resources`

joshuaspence:
Use the `hiera_hash` function together with `create_resources` for the `$consul::acls`, `$consul::checks`, `$consul::services` and `$consul::watches` parameters. This allows these resources to be defined in multiple levels of the Hiera hierachy.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -233,19 +233,19 @@ class consul (
   }
 
   if $services {
-    create_resources(consul::service, $services)
+    create_resources(consul::service, hiera_hash('consul::services', $services))
   }
 
   if $watches {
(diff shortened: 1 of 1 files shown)


solarkennedy:
Sorry, I don't think modules should be calling hiera directly like this. Using `hiera_hash` isn't necessarily the behavior that users want, it should be opt-in.

However if you need this behavior, you can use it like this:
```puppet
class { 'consul': 
  services => hiera_hash('consul::services'),
  ...
}
```
And get any special hiera behavior (optional hierarchy levels, etc), but the module it self I think should be "naive" and just accept input parameters.

PR #288, opened 2016-10-11
Prepared Queries

djtaylor:
Add a new provider for prepared queries API

PR #296, opened 2016-11-09
add support for prepared query templates

adamcstephens:
Expanding on https://github.com/solarkennedy/puppet-consul/pull/288 and the good work by @djtaylor, add support for prepared query templates to the custom type.

### DIFF
--- a/lib/puppet/provider/consul_prepared_query/default.rb
+++ b/lib/puppet/provider/consul_prepared_query/default.rb
@@ -158,57 +158,44 @@ Puppet::Type.type(:consul_prepared_query).provide(
     hostname = @resource[:hostname]
     protocol = @resource[:protocol]
     tries = @resource[:api_tries]
+    template = @resource[:template]
     template_regexp = @resource[:template_regexp]
     template_type = @resource[:template_type]
     prepared_query = self.get_resource(name, port, hostname, protocol, tries)
+    query_data = {
+      "Name"    => "#{name}",
+      "Token"   => "#{token}",
+      "Service" => {
+        "Service"     => "#{service_name}",
+        "Failover"    => {
+          "NearestN"    => service_failover_n,
+          "Datacenters" => service_failover_dcs,
+        },
+        "OnlyPassing" => service_only_passing,
+        "Tags"        => service_tags,
+      },
+      "DNS"    => {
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/consul_prepared_query/default.rb
```diff
@@ -179,7 +181,12 @@ def flush
         },
         "DNS"    => {
           "TTL" => "#{ttl}s"
-        }})
+        },
```
solarkennedy: Are you sure this should be part of this json should be included no matter what?

What about users of consul before 6.4? What if they don't want to make a template?

lib/puppet/type/consul_prepared_query.rb
```diff
@@ -108,4 +108,18 @@
       raise ArgumentError, "Number of API tries must be a number" if not value.is_a?(Integer)
     end
   end
+
+  newparam(:template_regexp) do
+    desc 'regexp for template'
+    validate do |value|
```
solarkennedy: Do you have to have to make this a required argument?

Can you default to nil so that users don't have to provide this all the time?

adamcstephens: Yeah, for sure. I'm somewhat new to ruby types so I'll read the docs a bit more and tweak this.

lib/puppet/provider/consul_prepared_query/default.rb
solarkennedy: Yes please.

lib/puppet/type/consul_prepared_query.rb
solarkennedy: Simply defaulting to "nil" or whatever and then not injecting template data will make this "automatically" backwards compatible.

adamcstephens:
I've updated this to add a template parameter, which allows 0.6.3 to work still since its defaulted to false.

lib/puppet/type/consul_prepared_query.rb
adamcstephens: Setting a default value to nil causes a failure to compile. Set them to some sane string defaults.

adamcstephens:
Any more feedback on this?

solarkennedy:
Nope! ty.

PR #299, opened 2016-11-28
Initial support for FreeBSD platform

olevole:


solarkennedy:
Can you also update the metadata.json?

olevole:
@solarkennedy Yes, I prefer to make a small portions of pull request for better code review ;) 
I need more time to check some part with FreeBSD (probably I need advice from you - I'll write you a mail after some testing). But I would be happy if you add these pull request, if there are no questions. I tested on FreeBSD 11.0-RELEASE and 12-CURRENT (aka HEAD), but I plan to move all my production with Consul to this module, so, FreeBSD 10.x  will also be tested soon.

solarkennedy:
Cool!

PR #303, opened 2016-12-17
Check $::consul_version more correctly

logic:
If we are using puppet-consul to download and install consul for us, initially `$::consul_version` will be undefined (because no executable could be found). This change makes the version check fail so we proceed with an installation (but without a puppet undefined variable error, which halts execution entirely, causing installation to fail).

There might be a better way to solve this, but it's been working for me. :)


solarkennedy:
You are correct, this is the proper way to deal with variables that may not exist. (esp with strict_variables on)

Issue #310, opened 2017-01-04
Annoying "defined 'protocol' as 'http' " for every consul_acl resource


mrwulf:
>Notice: /Stage[main]/Profiles::Consul/Consul_acl[Management]/protocol: defined 'protocol' as 'http'
Notice: /Stage[main]/Profiles::Consul/Consul_acl[Client API Token]/protocol: defined 'protocol' as 'http'
Notice: /Stage[main]/Profiles::Consul/Consul_acl[anonymous]/protocol: defined 'protocol' as 'http'
Notice: /Stage[main]/Profiles::Consul/Consul_acl[QA Admin]/protocol: defined 'protocol' as 'http'
Notice: Finished catalog run in 0.64 seconds

PR #312, opened 2017-01-10
Changes every run

mrwulf:
This should take care of issue #310 . I'm not quite sure why the normal puppet wrapping code isn't just 'making this work', but this change will do the trick.


solarkennedy:
Hmm. The provider code of this module is certainly my weak spot.

Issue #290, opened 2016-10-24
Consul binary not found when puppet agent is daemonized


mspaulding06:
Sometimes the consul binary cannot be found when resolving facts on a system.  This results in information like the consul_version fact not being available to puppet modules.  It turns out that the problem happens because the default install location of the consul binary symlink is `/usr/local/bin`.  When the puppet agent is daemonized on system start it is run by the init process which does not have `/usr/local/bin` in its path.  If puppet is run by a user with `sudo puppet agent -t` this all works fine because typically `/usr/local/bin` is in the path, but all automated runs will fail to resolve these facts.

There are a couple of possible solutions to this problem.  I didn't submit a PR because I figured it would be better to discuss what the right fix is first.  Solutions are:
1. Modify the fact so that it points to the absolute path of the consul binary (`/usr/local/bin/consul`).
2. Modify `/etc/sysconfig/puppet` on all systems to include `/usr/local/bin` in the path.

The first solution seems like it makes sense, but the problem is that the `bindir` variable in the consul module is configurable, and there is no way to read `bindir` from a custom fact.  The second solution requires a global modification to the system in order for this to work.  I guess one possible solution is it make it so that the `bindir` location is not configurable and symlink into a location that is guaranteed to exist (`/usr/bin`) instead.

solarkennedy:
Blarg :(

`/usr/local/bin` "feels" like the right place to put a randomly downloaded file. 

As a work around you could change the bin dir:
https://github.com/solarkennedy/puppet-consul/blob/605f5a705187f5783df20a67d0eb70b978a3ae09/manifests/init.pp#L16

Also we could "hack" the path in the fact itself.

mspaulding06:
I agree that `/usr/local/bin` "feels" like the right place.  It just, sadly, doesn't seem to work.  And I assume I'm not the only person affected by this.  It would be unfortunate to require all users of the module to set `bin_dir` in order to get facts to resolve.

When you say "hack" the path, are you referring to setting the fact's binary to use an absolute path like `/usr/local/bin/consul`?  That solution would be fine with me, just that the `bin_dir` could not be changed then.

solarkennedy:
This is the first time this issue has been raised, but yes you are probably not the only person to be affected.

When I say "hack" I mean literally adjust the path in the environment during execution of the fact to include `/usr/local/bin/` if it ins't in there.

mspaulding06:
[...]

PR #313, opened 2017-01-10
Ensure /usr/local/bin is in the path for consul binary

mspaulding06:
Checks the current path prior to running the consul binary.

* If no PATH is set, then a default value is set that includes `/usr/local/bin`
* If PATH already includes `/usr/local/bin` then no change is made to the path
* Otherwise, `/usr/local/bin` is added to the PATH environment variable

Tested change on CentOS 6.8 and verified that it fixes the problem I had encountered documented in #290

### DIFF
--- a/lib/facter/consul_version.rb
+++ b/lib/facter/consul_version.rb
@@ -3,10 +3,15 @@
 Facter.add(:consul_version) do
   confine :kernel => 'Linux'
   setcode do
+    original_path = ENV['PATH']
+    path = ENV.fetch('PATH') { '/bin:/usr/bin:/usr/local/bin' }
+    ENV['PATH'] = path + ':/usr/local/bin'
     begin
       Facter::Util::Resolution.exec('consul --version 2> /dev/null').lines.first.split[1].tr('v','')
     rescue
       nil
+    ensure
+      ENV['PATH'] = original_path
     end
   end
 end


solarkennedy:
I would like the path to be set back after. The reasoning here is that I don't want weird bugs to crop up because the path variable has changed during the puppet run when operators are not expecting that.

mspaulding06:
Alright, I made the change so that the `PATH` environment variable is restored after consul is run.  Good catch!

solarkennedy:
Can you double check the tests? It looks like they are failing on something fact-related.

solarkennedy:
Can you try rebasing on master to get current working tests?

xavvo:
Hi Matt, maybe you should save the result of the exec over the last line restoring the ENV :)

mspaulding06:
Thanks for the tip.  I've fixed the code, tests are passing.

lib/facter/consul_version.rb
```diff
@@ -4,7 +4,12 @@
...
     begin
-      Facter::Util::Resolution.exec('consul --version 2> /dev/null').lines.first.split[1].tr('v','')
+      tmp = ENV['PATH']
+      path = ENV.fetch('PATH') { '/bin:/usr/bin:/usr/local/bin' }
+      ENV['PATH'] = path + ':/usr/local/bin' unless path =~ /(:|^)\/usr\/local\/bin\/?(:|$)/
+      version = Facter::Util::Resolution.exec('consul --version 2> /dev/null').lines.first.split[1].tr('v','')
+      ENV['PATH'] = tmp
+      version
     rescue
       nil
```
solarkennedy: If the `exec` goes wrong, we need to set the path back.

lib/facter/consul_version.rb
```diff
@@ -4,7 +4,12 @@
   confine :kernel => 'Linux'
   setcode do
     begin
-      Facter::Util::Resolution.exec('consul --version 2> /dev/null').lines.first.split[1].tr('v','')
+      tmp = ENV['PATH']
+      path = ENV.fetch('PATH') { '/bin:/usr/bin:/usr/local/bin' }
+      ENV['PATH'] = path + ':/usr/local/bin' unless path =~ /(:|^)\/usr\/local\/bin\/?(:|$)/
```
solarkennedy: I think this is overkill. It doesn't hurt to append the path even if it is already there.
Just save the path, append the path, exec, reset path

lib/facter/consul_version.rb
```diff
@@ -4,7 +4,12 @@
   confine :kernel => 'Linux'
   setcode do
     begin
-      Facter::Util::Resolution.exec('consul --version 2> /dev/null').lines.first.split[1].tr('v','')
+      tmp = ENV['PATH']
```
solarkennedy: how about `original_path` instead of `tmp`?

solarkennedy:
Ok here I go!

mspaulding06:
Thank you!

PR #318, opened 2017-02-10
Keep the quotes for integers in "tags" and "node_meta" keys

### LINKED ISSUE #283: Quoted integer (string) to integer is breaking tags
In https://github.com/solarkennedy/puppet-consul/issues/119 quoted things that look like integers are explicitly turned into integers.

This breaks if you want a tag such as `"tags": ["106"]` as now you turn this into `"tags": [106]` and `consul` will no longer start, giving the error:
`* 'Tags[0]' expected type 'string', got unconvertible type 'float64'`.

Would it be possible just target areas of the config where you know the value needs to be an int instead of blanket changing them?

Thanks.

mmickan:
Without this change, integer values under tags and node_meta keys are not
quoted.  This results in errors from Consul such as:

* 'Tags[0]' expected type 'string', got unconvertible type 'float64'
* 'node_meta[cpu_count]' expected type 'string', got unconvertible type 'float64'

Fixes #283

### DIFF
--- a/lib/puppet/parser/functions/consul_sorted_json.rb
+++ b/lib/puppet/parser/functions/consul_sorted_json.rb
@@ -4,20 +4,23 @@ module JSON
   class << self
     @@loop = 0
 
-    def sorted_generate(obj)
+    def sorted_generate(obj, quoted=false)
       case obj
         when NilClass, :undef, Integer, Float, TrueClass, FalseClass, String
-          return simple_generate(obj)
+          return simple_generate(obj, quoted)
         when Array
           arrayRet = []
           obj.each do |a|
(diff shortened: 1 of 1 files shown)


solarkennedy:
Can you add a test to ensure this continues to work over time?

jardleex:
Hey there, is there a good chance to get this one fixed? We're moving more and more into Consul based on this module and I'm getting more use cases where these metadata would be handy.

bastelfreak:
@jardleex this PR is lacking tests and needs another rebase. Are you interested in picking this work up and continue it?

jardleex:
As much as I like to, the lack of any knowledge in Ruby and the tests to provide prevent me from doing so.

PR #325, opened 2017-03-01
Remove unnecessary .gitkeep files in lib/

cosmopetrich:
These get copied to puppet agent instances during pluginsync. That's not the end of the world, but it's not ideal either.

E.g.

```
Notice: /File[/opt/puppetlabs/puppet/cache/lib/puppet/provider/.gitkeep]/ensure: defined content as '{md5}d41d8cd98f00b204e9800998ecf8427e'
[...]
Notice: /File[/opt/puppetlabs/puppet/cache/lib/puppet/type/.gitkeep]/ensure: defined content as '{md5}d41d8cd98f00b204e9800998ecf8427e'
```


cosmopetrich:
I've left the `.gitkeep` files in templates/ and files/ alone since they're not getting in the way.

PR #334, opened 2017-04-13
Test changes for property

jk2l:
consul_key_value currently don't update if there is changes in value because it is set to param instead of property

also protocol has no state within this type, so it shouldn't be property as well

### DIFF
--- a/lib/puppet/type/consul_key_value.rb
+++ b/lib/puppet/type/consul_key_value.rb
@@ -20,7 +20,7 @@ Puppet::Type.newtype(:consul_key_value) do
     defaultto 0
   end
 
-  newparam(:value) do
+  newproperty(:value) do
     desc 'The key value string'
     validate do |value|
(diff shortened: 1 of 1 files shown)


jk2l:
btw i tested on my local environment, and it works. test case run fine on my local machine. 

p.s. your acceptence testing doesn't seem to work properly. it seem it hit this issue https://github.com/voxpupuli/puppet-archive/issues/225 so tested the module change directly on real server instead

nicholaspearson:
@solarkennedy @jk2l Is this resolved? If so, can we please merge this as this PR fixes an issue we are hitting.

jk2l:
tbh i am not sure did the author even read it...

solarkennedy:
I saw that the tests were failing and assumed you would be iterating.

I've rerun the tests and looks like it was something transient. Merging!

PR #341, opened 2017-05-19
support other unspecified RedHat variants

cspargo:
This allows the module to be used on other RedHat derivatives that are not specifically specified as supported. I have a proprietary variant, that I would like to use this module with, without specifically listing it as supported.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -61,17 +61,25 @@ class consul::params {
     } else {
       $init_style = 'systemd'
     }
-  } elsif $::operatingsystem =~ /Scientific|CentOS|RedHat|OracleLinux/ {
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -95,6 +95,12 @@
     $init_style = 'redhat'
   } elsif $::operatingsystem == 'FreeBSD' {
     $init_style = 'freebsd'
+  } elsif $::osfamily == 'RedHat' {
```
solarkennedy: Can you make this replace line 64?

cspargo:
updated to use osfamily fact for all redhat varieties, and added osfamily fact to the relevant tests

PR #345, opened 2017-06-06
Change http- options to rpc- ones

rawleto:
Hi, I've got an error with Ubuntu 14.04 and Consul v0.7.4:
```
~# consul reload -http-addr=localhost:8500
flag provided but not defined: -http-addr
Usage: consul reload

  Causes the agent to reload configurations. This can be used instead
  of sending the SIGHUP signal to the agent.

Options:

  -rpc-addr=127.0.0.1:8400 RPC address of the Consul agent.
```
So the pull makes reload_command option adjustable.


solarkennedy:
This version of puppet-consul targets 0.7 and up, please instead simply adjust the command to match the new required argument (`-rpc-addr`)

rawleto:
solarkennedy, thanks for the fast answer! I've edited options with sth like s/http/rpc/g and removed firstly created option.

rawleto:
Yeah, with the latest edits it works well on my boxes:
```
Info: Class[Consul::Run_service]: Scheduling refresh of Service[consul]
Notice: /Stage[main]/Consul::Run_service/Service[consul]: Triggered 'refresh' from 1 events
Notice: Applied catalog in 51.98 seconds
```

rawleto:
Some tests are broken cause http-port is 8500 but rpc-port is 8400.
I've adjusted ports in tests.

rawleto:
Yay! Tests are green again. Kennedy, I've done please review.

rawleto:
Updated to Consul v0.8.3 and the problem has gone. Could you just update default Consul version to 0.8.3 from 0.7.4 in the consul::params from the module > 3.0.0?

solarkennedy:
Well... that is because consul 0.8 removed rpc-add *entirely*!
https://www.consul.io/docs/upgrade-specific.html#consul-0-8-0

I'm not ready to make such a breaking change on 0.8 yet.

solarkennedy:
Whoops, I mean, master *is* targeting 0.8

So... I guess we should remove the rpc-addr options entirely!!

Issue #269, opened 2016-06-28
Add proxy support


walkamongus:
Consul is unable to download from behind a proxy. Proxy parameters should be added and then passed down into the archive resources to enable this.

solarkennedy:
I don't think this is really necessary or advisable. This module uses
https://github.com/voxpupuli/puppet-archive#archive

You could either set your proxy at the system level (setting the `http_proxy` variable):
http://www.cyberciti.biz/faq/linux-unix-set-proxy-environment-variable/

Or if you really must do it in puppet, you should do it in such a way so that it applies to all modules, not just this one:

```puppet
Archive {
  proxy_server => 'foo',
}
```
https://docs.puppet.com/puppet/latest/reference/lang_defaults.html

walkamongus:
Hi,

Thanks for the note on resource defaults.

I definitely wouldn't want to set a system-wide proxy as mentioned in the article you linked to. It seems to assume interactive and/or login shells. Also, a system-wide proxy setting will break connections to the local LAN unless your proxy is configured to do some sort of intelligent hairpinning for local LAN connections.

While the resource default works, I think there is a compelling argument against it. The resultant code:
```
Archive {
  proxy_server => 'foo',
}

class {'consul':
  param => 'bar',
  param => 'baz',
}
```
The Archive resource default does not appear to be related to anything. It only exists because of the knowledge of the internal implementation of the Consul module. I think having to do this is actually an argument **for** exposing a proxy setting in the Consul module's public interface. This is a pretty common situation with module dependencies -- very often you need to expose some of the interface of the internal modules in the interface of the containing module so that the behavior of the modules used internally can be changed.

Using the resource default could also potentially affect resources it shouldn't (Archive resource on local LAN that shouldn't go through proxy) if they happen to fall within its scope as well.

solarkennedy:
Alright you have convinced me, but only as much to leave the issue open for you or someone else to make the PR. I just worry about exposing every single option to everything in a puppet module.

I understand that there is the use case for only using an http proxy with this one puppet module, but imo a proxy setting *is* a system-wide thing. (You use the no_proxy var to specify the names/ips to not use the proxy for)

walkamongus:
I agree it would be nice to be able to set this system-wide but it's up to each program as to whether it respects the `*_proxy` environment variables (and I've run into some that don't).

[...]

PR #351, opened 2017-07-07
allow to specify a proxy server for package downloads

xavvo:
sometimes we need to specify a proxy to download the consul package. So we can just pass the corresponding puppet-archive parameter.


solarkennedy:
Is the use-case that only this *specific* download needs use a proxy server, but nothing else in the catalog?

Can you read https://github.com/solarkennedy/puppet-consul/issues/269 first and see if it is relevant?

xavvo:
Thank you for the issue pointer. I also think it is cleaner to specify the proxy to the consul module instead of the archive class, because this would create a hidden dependency that breaks if you happen to use something different to download the consul package.

And yes, we do use the archive module for other use cases that do not require the proxy.

manifests/install.pp
```diff
@@ -37,6 +37,7 @@
       -> archive { "${install_path}/consul-${consul::version}.${consul::download_extension}":
         ensure       => present,
         source       => $::consul::real_download_url,
+        proxy_server => $consul::proxy_server,
```
solarkennedy: can you add the leading `$::` for consistency ?

solarkennedy:
Resolves #269, cc @walkamongus

walkamongus:
@solarkennedy thanks for the followup!

PR #352, opened 2017-07-14
Added windows support

### LINKED ISSUE #195: Add Support for Windows
Adding this issue for tracking, my company will be working on providing this support as we have hundreds of windows boxes that will need this support.

krjensen:
This is an updated attempt to add windows support, it is based on the old branch `feature/windows_support` but has been updates and tested on server 2016.

Could we get this merged and mark windows support as experimental?

Fixes #195


solarkennedy:
Certainly! Can you rebase against current master first to resolve the merge conflicts?

krjensen:
Conflict should be solved now, I have simplified the logic around init_style for windows, so it reuse existing logic around `$init_style = 'unmanaged'`

manifests/install.pp
```diff
@@ -4,16 +4,31 @@
 #
 class consul::install {
 
+  case $::operatingsystem {
+    'windows': {
+      $binary_name = 'consul.exe'
+      $binary_mode = '0775'
```
solarkennedy: why does windows need different permissions here?

If we already know what the group should be, I'm ok with making 775 being the default for both oses. Does 555 not work for running windows binaries?

krjensen:
Let me do another test tomorrow and I'll let you know.

manifests/install.pp
krjensen: I just ran a test on the, and the binary_mode can be set to 755, but 555 does not work on windows 2016. I have pushed a version where I set it to 755 for both data_dir and binary

robklg:
Hey, I was just looking into adding windows support to this puppet module. I see this merge request was about to get merged, but hasn't yet?

Is there something I can do?
Does it needs some more testing first? We use Windows Server 2012.

solarkennedy:
The main blocker is that this PR installs nssm on behalf of the user, which I don't think this module should do.

solarkennedy:
OMG it looks like all my review comments are still pending because I didn't hit the "submit review" button... :(

robklg:
@solarkennedy yeah I understand your reluctance. I do agree however that NSSM is the best option. But it would also be fine to me if the puppet-consul module would only manage the config and not the installation. Then we can do the NSSM part in our profile.

solarkennedy:
I agree that would be a good compromise. For the same reason this puppet module doesn't install curl or unzip (as much as other people want it to), I don't think it is sustainable. If every puppet module that need to use the unzip binary tried to manage that package there would be lots of conflicts.

If the consul module tries to install NSSM, it would conflict with any other puppet module that thinks it should install NSSM too. NSSM should "just" be its own puppet module, puppet-consul should "just" use that or assume it is available.

iwagnerclgx:
Any thoughts on proceeding? I don't mind taking a stab at rebasing @krjensen's PR into the latest master and cleaning up the concerns raised here.

solarkennedy:
Go for it

solarkennedy:
Closed by #403

PR #355, opened 2017-07-20
add note in README

AntonTumasov:
Just to prevent people from failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ Optionally, you may supply an `acl_api_token`.  This will allow you to create
 ACLs if the anonymous token doesn't permit ACL changes (which is likely).
 The api token may be the master token, another management token, or any
 client token with sufficient privileges.
+### NOTE. If you're already using ACL tokens you should consider not to use HCL due to library using JSON.parse function to create/update tokens
 
 ## Prepared Queries and Prepared Query Templates
 


README.md
```diff
@@ -204,6 +204,7 @@ Optionally, you may supply an `acl_api_token`.  This will allow you to create
 ACLs if the anonymous token doesn't permit ACL changes (which is likely).
 The api token may be the master token, another management token, or any
 client token with sufficient privileges.
+### NOTE. If you're already using ACL tokens you should consider not to use HCL due to library using JSON.parse function to create/update tokens
```
solarkennedy: This reads a little funny. 

What does HCL have to do with this? And which json library, the one that puppet-consul uses?

AntonTumasov: It's related to those, who faced the same problem as we did. It doesn't work with beforehand set up ACL rules in Hashicorp DSL. And yes, JSON from your Ruby code.

solarkennedy: What was the actual result? Did puppet overwrite the tokens that were created with HCL?

AntonTumasov: Nope, it didn't. It just failed with parsing. That's why I added this note.

solarkennedy: How about

```
### NOTE: This module currently cannot parse ACL tokens generated through means other than this module. Don't mix Puppet and Non-puppet ACLs for best results!
```

AntonTumasov: If it suits you, do so.

solarkennedy: Will do.

PR #357, opened 2017-08-07
added docker support

Justin-DynamicD:
added two new install_methods:  auto and docker (neither set to default).

docker now uses the official docker container to deploy but still calls config_hash and mounts the config_dir for consistency.  Image can be chosen, but network and certain key variables are hardcoded per consul recomendations for use in a container by the host.  See: https://hub.docker.com/_/consul/

auto will look for the [docker0] network interface.  If found, it uses docker, on failure it falls back to the url default.

The default remains at "url", disabling this feature.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -84,10 +84,10 @@ define consul::check(
   consul_validate_checks($check_hash[check])
 
   $escaped_id = regsubst($id,'\/','_','G')
-  
+
   file { "check_${escaped_id} with consul accounts" :
-    path    => "${consul::config_dir}/check_${escaped_id}.json",
     ensure  => $ensure,
+    path    => "${consul::config_dir}/check_${escaped_id}.json",
     owner   => $::consul::user_real,
     group   => $::consul::group_real,
     mode    => $::consul::config_mode,
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -85,26 +85,15 @@ define consul::check(
 
   $escaped_id = regsubst($id,'\/','_','G')
   
-  if $::consul::selected_install_method != 'docker' {
-    file { "check_${escaped_id} with consul accounts" :
-      path    => "${consul::config_dir}/check_${escaped_id}.json",
-      ensure  => $ensure,
-      owner   => $::consul::user,
-      group   => $::consul::group,
-      mode    => $::consul::config_mode,
-      content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
-      notify  => Class['consul::reload_service'],
-      require => File[$::consul::config_dir],
-    }
+  file { "check_${escaped_id} with consul accounts" :
+    path    => "${consul::config_dir}/check_${escaped_id}.json",
+    ensure  => $ensure,
+    owner   => $::consul::user,
+    group   => $::consul::group,
+    mode    => $::consul::config_mode,
+    content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
+    notify  => Class['consul::reload_service'],
+    require => File[$::consul::config_dir],
   }
-  else {
-    file { "check_${escaped_id} without consul accounts" :
-      path    => "${consul::config_dir}/check_${escaped_id}.json",
-      ensure  => $ensure,
-      mode    => $::consul::config_mode,
-      content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
-      notify  => Class['consul::reload_service'],
-      require => File[$::consul::config_dir],
-    }
-  }
+
 }
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -85,11 +85,22 @@ define consul::check(
 
   $escaped_id = regsubst($id,'\/','_','G')
   File[$::consul::config_dir]
-  -> file { "${consul::config_dir}/check_${escaped_id}.json":
+  -> file { "check_${escaped_id} with consul accounts":
+    path    => "${consul::config_dir}/check_${escaped_id}.json",
     ensure  => $ensure,
     owner   => $::consul::user,
     group   => $::consul::group,
     mode    => $::consul::config_mode,
     content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
-  } ~> Class['consul::reload_service']
+    notify  => Class['consul::reload_service'],
+    unless  => "/usr/bin/test -e ${consul::config_dir}/docker_used",
+  }
+  -> file { "check_${escaped_id} without consul accounts":
+    path    => "${consul::config_dir}/check_${escaped_id}.json",
+    ensure  => $ensure,
+    mode    => $::consul::config_mode,
+    content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
+    notify  => Class['consul::reload_service'],
+    onlyif => "/usr/bin/test -e ${consul::config_dir}/docker_used",
+  }
 }
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -84,16 +84,13 @@ define consul::check(
   consul_validate_checks($check_hash[check])
 
   $escaped_id = regsubst($id,'\/','_','G')
-
-  file { "check_${escaped_id} with consul accounts" :
+  File[$::consul::config_dir]
+  -> file { "${consul::config_dir}/check_${escaped_id}.json" :
     ensure  => $ensure,
-    path    => "${consul::config_dir}/check_${escaped_id}.json",
     owner   => $::consul::user_real,
     group   => $::consul::group_real,
     mode    => $::consul::config_mode,
     content => consul_sorted_json($check_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
-    notify  => Class['consul::reload_service'],
-    require => File[$::consul::config_dir],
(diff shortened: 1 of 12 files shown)


manifests/params.pp
```diff
@@ -40,7 +41,7 @@
   $ui_package_ensure     = 'latest'
   $ui_package_name       = 'consul_ui'
   $user                  = 'consul'
```
solarkennedy: Many of the complexity in this PR revolves around not using the `user` and `group` variables.
Can you simply set them to be `undef` in params.pp if the install method is docker instead? (then we won't need case statements throughout the codebase, and instead only have them in one place.

solarkennedy: Same with manage_group and manage_user

manifests/init.pp
```diff
@@ -52,9 +55,12 @@
 # [*init_style*]
 #   What style of init system your system uses. Set to 'unmanaged' to disable
 #   managing init system files for the consul service entirely.
+#   This is ignored when install_method == 'docker'
 #
 # [*install_method*]
-#   Valid strings: `package` - install via system package
+#   Valid strings: `auto`    - use docker if available, fallback to url if not
```
solarkennedy: I'm going to veto this.
I think it adds complexity and surprise.
Just have a default method in the module and make users set it "automatically" themselves with their own puppet code.

manifests/params.pp
solarkennedy: Also init_style

manifests/run_service.pp
```diff
@@ -24,14 +24,47 @@
...
+        command => $command
+      }
+  }
+
   if $::consul::join_wan {
     exec { 'join consul wan':
       cwd       => $::consul::config_dir,
       path      => [$::consul::bin_dir,'/bin','/usr/bin'],
       command   => "consul join -wan ${consul::join_wan}",
+      onlyif    => "/usr/bin/test ! -e ${consul::config_dir}/docker_used",
```
solarkennedy: Can you get away with a normal if statement here instead of the `docker_used` workaround?

manifests/params.pp
Justin-DynamicD: Not sure why I didn't think of doing that in the prams vs trying to remap it in the init.pp ... lost in the moment.  But yes:  I deliberately attempted to clear user statements as they aren't needed in the container.

I'll rework it tonight/next few days.  Thanks for the input.

manifests/run_service.pp
Justin-DynamicD: probably need to assemble $command variable then add a prefix if install_method == docker.  If you look close, in order to run consul in the container I need to call it slightly differently (prefix it with "docker exec -t consul <...>").  So ... yeah if we "veto auto" we can get away from that.

manifests/init.pp
Justin-DynamicD: oh fine, take the fun out of it!

Justin-DynamicD:
changes made including removal of auto.  Had to make some "minor tweaks" as you can't rewrite variables in puppet but should be closer inline to what you're looking for.

Justin-DynamicD:
took way to long to get things to pass, but there you go.

Justin-DynamicD:
Running internally at the company against Ubuntu 14, 16, and CentOS 7.9 in both docker and non-docker configs.  Found a minor bug where a container env variable was writing a new local.json.  Removing to allow the user complete control over config through the module.

manifests/install.pp
```diff
@@ -4,16 +4,19 @@
 #
 class consul::install {
 
-  if $::consul::data_dir {
-    file { $::consul::data_dir:
+  if ($::consul::data_dir) and ($::consul::install_method != 'docker') {
```
solarkennedy: Why is there no data_dir? I would prefer to have a permanently mounted volume outside the container. That way if I have to nuke docker for some reason, puppet can reproduce the image and run and all it has to do is pickup the data on disk

manifests/params.pp
```diff
@@ -11,7 +11,8 @@
   $config_defaults       = {}
   $config_dir            = '/etc/consul'
   $config_hash           = {}
-  $config_mode           = '0660'
+  $config_mode           = '0664'
```
solarkennedy: Can you leave it at 660 to continue to allow potentially sensitive configs unreadable by random users of the system or hacked nobodys or whatever?

manifests/reload_service.pp
```diff
@@ -19,12 +19,16 @@
       $http_addr = $::consul::http_addr
     }
 
+    case $::consul::install_method {
+      'docker': { $command = "docker exec -t consul consul reload -http-addr=${http_addr}:${consul::http_port}" }
```
solarkennedy: does docker exec need a `-t` here? shouldn't need a tty right?

manifests/run_service.pp
```diff
@@ -24,14 +24,47 @@
...
+    if $server_mode {
+      $env = [ '\'CONSUL_ALLOW_PRIVILEGED_PORTS=\'' ]
+      $docker_command = 'agent -server'
+    }
+    else {
+      $env = undef
+      $docker_command = 'agent'
+    }
+
+    # Docker Install
```
solarkennedy: remove the comment, doesn't really provide value here

manifests/check.pp
```diff
@@ -85,11 +85,12 @@
 
   $escaped_id = regsubst($id,'\/','_','G')
   File[$::consul::config_dir]
-  -> file { "${consul::config_dir}/check_${escaped_id}.json":
+  -> file { "${consul::config_dir}/check_${escaped_id}.json" :
```
solarkennedy: unless this is a new change in the style, can you remove the space before the colon?

manifests/run_service.pp
```diff
@@ -24,14 +24,47 @@
...
+    docker::run { 'consul' :
```
solarkennedy: space before the colon

manifests/config.pp
```diff
@@ -93,25 +93,26 @@
...
     ensure  => 'directory',
-    owner   => $::consul::user,
-    group   => $::consul::group,
+    owner   => $::consul::user_real,
+    group   => $::consul::group_real,
     purge   => $purge,
     recurse => $purge,
   }
-  -> file { 'consul config.json':
+  -> file { 'consul config.json' :
```
solarkennedy: space

manifests/install.pp
```diff
@@ -4,16 +4,19 @@
...
+    file { $::consul::data_dir :
```
solarkennedy: space

manifests/watch.pp
```diff
@@ -134,11 +134,12 @@
   }
 
   File[$::consul::config_dir]
-  -> file { "${consul::config_dir}/watch_${id}.json":
+  -> file { "${consul::config_dir}/watch_${id}.json" :
```
solarkennedy: space

manifests/install.pp
Justin-DynamicD: Mostly because of the nature of consul+puppet itself makes it unnecessary.  In a production environment you should have 3/5/n masters so all info will replicate, and with a puppet config all checks and services will populate on first run so it's simply unnecessary to store that information beyond the life-cycle of the container (remember volumes will persists in upgrades, just not with you deliberately rm the container)

manifests/params.pp
Justin-DynamicD: this will break container volume mounts as the container itself does not share the users on the host, and thus counts as "other users".  It was set as read-only to prevent tweaking.  Keep in mind that in a container scenario the surface area of attack will be the containers themselves (that hosts all open ports).  I can add a "config_mode_real" so it can be flipped back and forth based on the deployment mode?

manifests/reload_service.pp
Justin-DynamicD: That's a good question.  I have it stuck in my head that's run into issues without the TTY enabled ... but honestly this could be based on earlier releases/info.  Ill give a test internally and see what's up with that.

manifests/check.pp
Justin-DynamicD: Formatting habbit brought on from a vscode plugin for puppet.  Can be pulled.

manifests/run_service.pp
Justin-DynamicD: ditto

manifests/config.pp
Justin-DynamicD: ditto

manifests/install.pp
Justin-DynamicD: ditto

manifests/watch.pp
Justin-DynamicD: ditto

Justin-DynamicD:
Spaces removed and data_dir added.

Justin-DynamicD:
This should be ready for another review.  Please let me know.

solarkennedy:
This is still on my radar and I will review. 

Can you confirm: are you running this code (in the PR) for real in your infrastructure? or is it a variant of it?

Justin-DynamicD:
This code is the code deployed at my company.  We are using a mix of Ubuntu (14 and 16) as well as a few CentOS boxes, and I deploy both Docker and URL mode so I have a pretty good cross section, but we are a small shop of only ~80 boxes on premise.  It's currently used in tandem with Registrator to do most of the service registration, but I do have a few specific service callouts being done.

solarkennedy:
Alright, hitting the button...

Justin-DynamicD:
Sweet.  Ill remove my branch and open a new one If i come up with more ideas.

PR #372, opened 2017-10-13
Validate config file before deploying

kasimon:
Use validate_cmd to check the config file for validity before actually installing it.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -106,13 +106,14 @@ class consul::config(
     recurse => $purge,
   }
   -> file { 'consul config.json':
-    ensure  => present,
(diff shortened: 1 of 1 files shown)


solarkennedy:
Hmm. This seems like a good idea. It does imply that the consul binary is installed and ready to go. Seems legit.

kasimon:
Yes, the binary has to be present, but it does not have to be set up in any way. The validation happens independently from the normal service setup.

Thanks for merging!

wyardley:
I know there is a compatibility matrix in the README, but this does break consul on versions where the version doesn't yet support `consul validate`. Maybe there should be a conditional based on consul version?

Issue #376, opened 2017-10-25
Validate command for config files doesn't work


joshuaspence:
With Consul 1.0, the change from #372 doesn't seem to work because the temporary config file doesn't end with `.json`.

```
       	Error: Execution of '/usr/local/bin/consul validate /etc/consul/config.json20171025-5833-1sgyj3d > /dev/null' returned 1: Config validation failed: Missing or invalid file extension for "/etc/consul/config.json20171025-5833-1sgyj3d". Please use ".json" or ".hcl".
       	Error: /Stage[main]/Consul::Config/File[consul config.json]/ensure: change from absent to present failed: Execution of '/usr/local/bin/consul validate /etc/consul/config.json20171025-5833-1sgyj3d > /dev/null' returned 1: Config validation failed: Missing or invalid file extension for "/etc/consul/config.json20171025-5833-1sgyj3d". Please use ".json" or ".hcl".
```

joshuaspence:
It looks like Consul version 1.0.1 addresses this problem, see hashicorp/consul#3620.

khdevel:
It is almost fixed... the only thing is missing is that the symbolic link still exist so puppet has problem during some second _config.json_ update
```
Error: Execution of '/usr/local/bin/consul validate $(ln -s /etc/consul/config.json20171114-19498-11xf5vn /tmp/consul-config.json && echo "/tmp/consul-config.json") > /dev/null' returned 1: ln: failed to create symbolic link '/tmp/consul-config.json': File exists
```
Please add some "rm like" command and I think it will be done here.

solarkennedy:
I just merged in a PR that does `ln -sf`

joshuaspence:
The symbolic link approach feels so dirty... Why not just use `-config-format=json`?

PR #377, opened 2017-10-27
Fix validate command for config files in consul 1.0.0

Faffnir:
The consul validate command has a problem with using files which do not end in .json or .hcl which is the fact when using validate_cmd which uses a temporary file and puppet adds a tmp-specific string. This solution is kind of an ugly workaround which produces a tmp link file with the correct name and leaves it in /tmp. After the check this link will become invalid and there are no security issues.

The consul maintainers are discussing about a better solution here https://github.com/hashicorp/consul/issues/3620 but we need a solution right now and that fixed it for us. This could maybe be removed after consul adds a fix for this issue.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -113,7 +113,7 @@ class consul::config(
     mode         => $::consul::config_mode,
     content      => consul_sorted_json($config_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
     require      => File[$::consul::config_dir],
-    validate_cmd => "${::consul::bin_dir}/consul validate % > /dev/null",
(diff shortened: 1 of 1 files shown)


Faffnir:
This would fix [https://github.com/solarkennedy/puppet-consul/issues/376](Issue 376)

solarkennedy:
I would also be OK if we gave up trying to do client-side validation like this.

Faffnir:
And you are actually right in suggesting this because I just found out that I introduced a nasty bug as it is now not possible to update the consul config without getting an error in puppet.
Should I just create a new PR with excluding the check totally @solarkennedy ?

solarkennedy:
Yea. I think it was a good thought, but a tad expensive and error prone.

Issue #374, opened 2017-10-23
Changing Consul kv values doesn't seem to have an effect


dannytrigo:
I have tried to test manually changing a kv value after it has been set by the puppet-consul module but it never seems to be changed back by Puppet.
I have added debug code and as far as I can tell, the new value is retrieved from Consul, but then it seems that the flush method is not being called by Puppet, even though the value is different.
Have others had more success with this?

solarkennedy:
It sounds like #363, can you try on master?

dannytrigo:
I tried on master and although better (flush now gets called even though I couldn't figure out why it wasn't before), it still seems to have problems.
Firstly, when the key exists already, it seems `@property_flush[:ensure]` is not set (since its ensure parameter is not changing, so I got:
```
Could not evaluate: ensure attribute is set to unexpected value:
```
So I removed that condition check.  Then I discovered that the value still isn't changed because it seems that in the comparison between `key_value[:value]` and `value` both variables are somehow the new value.

So I removed that check against the current value and it now seems to be working.  Flush only seems to be called when a parameter has changed, so I'm not sure the comparison is actually needed.  Without, puppet is only updating the keys that need updating.

dannytrigo:
@solarkennedy - any opinions on that? Should I open a PR?

solarkennedy:
Yes please. I'll ask @houtmanj to review it (the provider was significantly refactored recently).

houtmanj:
@dannytrigo @solarkennedy I prefer we find out why the underlying assumptions do not behave properly. 
According to the docs the create/delete methods should allways be called, same as the flush method. 
It will very likely also have effect on the major refactoring still pending review and the other providers. 

I currently only have access to a browser and cannot debug the issue any further untill the end of the week. 

@dannytrigo what puppet version are you running? It seems your case is easily reproduced which is nice.

dannytrigo:
Are create/delete always called even if the resource does not need creating, only altering?

I'm using puppet 3.8.7 which is very old and may explain the difference in behaviour.  Can you point me to the docs because I've had a lot of trouble locating good documentation on writing provider plugins.

houtmanj:
Here are some of the resources I found:
- https://puppet.com/docs/puppet/4.10/complete_resource_example.html
- http://garylarizza.com/blog/2013/12/15/seriously-what-is-this-provider-doing/

Turns out I was wrong, the flush method indeed only gets called when there is a change. 
[...]

PR #378, opened 2017-10-30
Fix updating of Consul KV store

dannytrigo:
It seems @property_flush[:ensure] is only set when creating/deleting but is nil otherwise
https://github.com/solarkennedy/puppet-consul/issues/374

### DIFF
--- a/lib/puppet/provider/consul_key_value/default.rb
+++ b/lib/puppet/provider/consul_key_value/default.rb
@@ -137,6 +137,10 @@ Puppet::Type.type(:consul_key_value).provide(
   end
 
   def flush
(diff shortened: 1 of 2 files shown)


solarkennedy:
Is `info` really a good level to log here? How about `debug`?
 @houtmanj can you review?

dannytrigo:
Updated as per feedback

solarkennedy:
I'm ready to merge if both of you are happy.

dannytrigo:
Ok for me. @houtmanj?

houtmanj:
looks good to me!

solarkennedy:
This is in 3.2.0

PR #380, opened 2017-11-14
Removing the consul config check

Faffnir:
Originally this config check wasn't working with the newest consul version and a workaround with saving it temporarily was introduced. When updating the configuration this check wasn't working either so it is safer to remove it and leave it to the developer to set the correct information.

The new consul config check has a posibility to circumvent the original problem that the configuration file wasn't ending in ".json" but this would also introduce checking before which consul version is running.

This PR is regarding: https://github.com/solarkennedy/puppet-consul/pull/377#event-1314047133

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -106,14 +106,13 @@ class consul::config(
     recurse => $purge,
   }
   -> file { 'consul config.json':
-    ensure       => present,
-    path         => "${consul::config_dir}/config.json",
-    owner        => $::consul::user_real,
-    group        => $::consul::group_real,
-    mode         => $::consul::config_mode,
-    content      => consul_sorted_json($config_hash, $::consul::pretty_config, $::consul::pretty_config_indent),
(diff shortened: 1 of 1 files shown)


wyardley:
I had some problems with this on earlier versions of Consul as well (which don't have the `consul validate` command), when working on updating the acceptance tests.
While the current version of the module officially doesn't support < 0.8.x according to the README, maybe a workaround would be to only do the validation with a conditional?

There might be some other ways to handle this as well, though all the ones I can think of seem kind of hackish.

Faffnir:
I removed it now because I thought if you deploy consul to several servers with puppet you should have tested it before anyways? (maybe already with the validate consul argument which is awesome for testing)
I am not sure if a successful puppet run should break because of a faulty consul config?

solarkennedy:
I'm going to go ahead and merge this. It seemed like a cool idea at the time, but it isn't super easy to support.

PR #381, opened 2017-11-14
Breaking: Update Puppet version to 4.7.1 and add data types

wyardley:
This PR:
- Updates minimum required Puppet version to 4.7.1
- Updates allowed / required versions for stdlib and archive modules
- Switches deprecated `validate_xx` calls with Puppet data types (data types could be added to other class params later).
- Removes a test case for a conditional that tested versions < Consul 0.4 (I left the conditional in place, but I think not testing for it is Ok, since this version of the module doesn't support that version).
- Switches Fixnum => Integer to remove a Ruby deprecation warning in ` lib/puppet/parser/functions/consul_sorted_json.rb`
- Switches test matrix to test with Ruby 2.1.9 and 2.4.2 on Puppet 4 and 5 (vs. older Puppet, older Ruby).
- Makes at least some improvements to acceptance tests, they run locally on Ubuntu 14 without Docker. I could work on some additional improvements, including getting it running in Docker, if you have Travis setup to work with that.

This would be a breaking / backwards incompatible version update.

Feedback or requested changes welcome. I purposely left the test changes for removing that hiera hack, and the acceptance test changes separate, but can easily squash if you want in one commit.

I removed Gemfile.lock because it's in .gitignore. But I can check in a current one based on this PR if that's preferred.

Assuming you are amenable to dropping Puppet 3.x etc. support, you could create a puppet3 branch from before these changes, though since it's EOL, I think it makes sense to drop support completely.


wyardley:
@solarkennedy for your reviewing consideration....

solarkennedy:
I'm in favor of this change, my only issue is with the 4.7 minimum. I'm having trouble finding what the current supported versions of the 4.x series line are. Do you know or can you link?

wyardley:
@solarkennedy I was told 4.7.1; that is what Vox Pupuli uses as the standard for lowest supported version, and it doesn't hurt the score rating on the Forge. Could doublecheck, but I think it is a safe lower end.

Rebased to fix the merge conflict

wyardley:
@solarkennedy I came across a discussion about this issue on IRC. 2016.4 is current Puppet Enterprise LTS release; 2016.4.5 ships with agent 4.10.1.
https://puppet.com/docs/pe/2016.4/install_what_and_where.html
Anecdotally, "4.7.1 was the first stable one that worked properly with datatypes" according to @bastelfreak.
If you look at https://forge.puppet.com/KyleAnderson/consul/scores
"Parser validation against Puppet 4.7 succeeds." [under compatibility].

I also asked someone on the modules team; he didn't have any specific reference.
puppetlabs-apt, as one example, has 4.7.0 as the floor.
https://github.com/puppetlabs/puppetlabs-apt/blob/master/metadata.json

bastelfreak:
Thanks for this awesome PR @wyardley!

@solarkennedy I talked a lot to the Puppet people to find the correct version for Vox Pupuli and Puppet modules. We agreed on 4.7.0 (or 4.7.1 this this was a bugfix release). There are a few modules that require an even newer version, for example 4.9 if we have data-in-modules, but  4.7.1 is the lowest we officially support in the community.

solarkennedy:
Alright I'm going to press it. Towards the future!

wyardley:
Thanks @solarkennedy!

PR #388, opened 2017-11-27
set sane default shell for consul user

bastelfreak:


manifests/install.pp
```diff
@@ -117,6 +117,7 @@
       ensure => 'present',
       system => true,
       groups => $consul::extra_groups,
+      shell  => '/sbin/nologin',
```
solarkennedy: I'm in favor of this, but '/sbin/nologin' is not a universal thing.

Can you add it to params so it can be changed on a per-platform basis?
Here is another example of someone doing this:
https://github.com/geoffgarside/puppet-redis/blob/f06c835c6316cb18823e64c758a0b7545418c862/manifests/params.pp

On my server (Ubuntu) it is /usr/sbin/nologin.

bastelfreak:
@solarkennedy can you recheck this again please?

PR #391, opened 2017-12-04
Do not unquote integers with a starting zero

### LINKED ISSUE #389: consul_sorted_json for octal notation of file modes
We are about to [base the puppet consul-template module](
https://github.com/claranet/puppet-consul_template/issues/100) upon consul_sorted_json.rb.

Now consul-template has a `template.perms` configuration setting, which configures filesystem permissions in ocatal notation eg. `0644` which is not a valid integer.

The string match in [consul_sorted_json.rb#85](https://github.com/solarkennedy/puppet-consul/blob/ad0bdb0562d68d106c3b38f4773ab820aa3c140d/lib/puppet/parser/functions/consul_sorted_json.rb#L85) does not take octal representation in account.

phaer:
Integers with leading zeros are invalid in JSON, so they should probably stay quoted in. (They would be valid Javascript, but JSON does not support octal notation as stated on http://json.org.

Fixes #389

### DIFF
--- a/lib/puppet/parser/functions/consul_sorted_json.rb
+++ b/lib/puppet/parser/functions/consul_sorted_json.rb
@@ -82,7 +82,7 @@ module JSON
         else
           # Should be a string
           # keep string integers unquoted
-          (obj =~ /\A[-]?\d+\z/) ? obj : obj.to_json
+          (obj =~ /\A[-]?[1-9]\d*\z/) ? obj : obj.to_json
       end
     end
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
Please add a test case for this. If we don't, then this particular functionality will be lost as consul_sorted_json changes.

solarkennedy:
@phaer does this also fix #283 ?

phaer:
Just added tests and fixed the regex to still unquote for the case where the number is just '0'. As far as I understand, it does *not* fix #283 which looks for a way to keep valid integers quoted. This PR is only about (invalid) integers with a leading zero.

Issue #396, opened 2017-12-20
Support for Amazon 2 OS 


vchan2002:
should be a fairly trivial change as the OS will be using systemd.... I'll try to put in a PR in the near future.

https://aws.amazon.com/about-aws/whats-new/2017/12/introducing-amazon-linux-2/

PR #397, opened 2017-12-20
support Amazon Linux 2

vchan2002:
solves #396


solarkennedy:
Can you update metadata.json?

Issue #384, opened 2017-11-21
Allow for different consul and consul-ui versions.


madelaney:
With version 0.9+ of consul, there is no analog UI package. The last UI package was 0.8.5. This means that if I try to use consul 1.0.0 (which was original intent), I cannot also setup the UI.

bp85:
can some one address this issue

rawleto:
@madelaney @bp85 using #398 you can install the latest Consul after removing ui_* parameters if any and adding a 'ui => true' option to the config_hash (please follow a link to the details: https://github.com/solarkennedy/puppet-consul#web-ui).

Issue #358, opened 2017-08-14
cannot update to consul 0.9.0 because of a 'web_ui.zip' package


khdevel:
The module wants to install a *web_ui* package
```
extract_path => "${install_path}/consul-${consul::version}_web_ui",
```
which, since v0.9.0, is not availabe in Hashi's repository.
Please perform the fix in this line (as I assume) [install.pp#L56](https://github.com/solarkennedy/puppet-consul/blob/master/manifests/install.pp#L56)

I could not find any place to make it possible to NOT install the *web_ui* package. Removed *ui_id* did not help in this case. After this Consuls's WebUI was not available.

khdevel:
I used a *ui => true* in my puppet's configuration and removed *ui_dir => ...* section an problem is partly solved. But I think that there should be some notification or part in the code which will not download a *web_ui.zip* package from Hashi's repository.

solarkennedy:
Hmm. With no ui package I guess we should get rid of it for >=0.9 entirely?

Do whatever you've got to do to get things working for you, then please evaluate if you can PR the change for everyone.

devcfgc:
@khdevel Did you manage to fix it? perhaps the ui package has been include somewhere else since 0.9.0.. .

boweeb:
My understanding is that the ui package doesn't exist anymore (>=`0.9.0`) because it's now integrated with consul itself.  To verify, I deployed consul normally without any `ui_*` settings then manually changed the systemd service file to add the `-ui` flag and the ui worked out of the box without installing anything extra.

boweeb:
So a reasonable way to deploy the UI is to just add `-ui` to `extra_options` and not set any `ui_*`.  Perhaps the PR should be more around user experience / documentation.

rawleto:
@khdevel please check #398 out with ui options in the config_hash (see https://github.com/solarkennedy/puppet-consul#web-ui).

Issue #246, opened 2016-03-22
package_name and ui_package_name can't be the same


berlincount:
We use a custom RPM with consul that has consul_ui inside already ... installing that right now is impossible, because that gives a duplicate Package[consul] resource.

split3:
Are you providing a `$consul::ui_dir` as that is the only way that package will get installed. If you don't provide that value the package for the UI won't get installed.  If your package already combines them try not providing the `$consul::ui_dir` and see if that works.

split3:
So looks like it wouldn't be possible to omit the `$consul::ui_dir` as that is need in the config hash for consul and is looked up that way:
[https://github.com/solarkennedy/puppet-consul/blob/master/manifests/install.pp#L49](url)

So in this case we may need to option to forcibly exclude trying to install the UI package in this particular case.

rawleto:
@berlincount with #398 you can try to install RPMed Consul adding 'ui => true' and 'ui_dir => some_dir' options to the config_hash (please follow a link to the details: https://github.com/solarkennedy/puppet-consul#web-ui).

berlincount:
Ah well, I've moved on from Puppet ...

PR #398, opened 2017-12-26
Added the ui parameter and removed ui_* ones

rawleto:
Hi, I did some changes that break backward compatibility but work with UI from Consul >=v0.9.0


solarkennedy:
Does this address https://github.com/solarkennedy/puppet-consul/issues/384 ?

Or https://github.com/solarkennedy/puppet-consul/issues/246 ?

rawleto:
Actually I thought about #358, but these issues suit too.

solarkennedy:
Ok. I'll merge this in, but I expect some complaints :)

Can you comment on each of those issues please with the current state of the art on how to deal with the UI after I merge?

PR #401, opened 2018-01-16
Removed command substitution in init script

tooooots:
This command substitution causes Consul to be launched as root instead as the non-privileged user.
(Tested on AmazonLinux1)


solarkennedy:
Honestly I'm a little confused about this. It looks very funny as a command substitution, makes me question how it ever worked?

But then it has the `&`, is it backgrounding too? Yuck.

PR #403, opened 2018-01-25
Added experimental Windows support

iwagnerclgx:
Here is a slightly modified version of #352.

- NSSM is expected to be installed prior, and the nssm executable passed as a parameter.
- Added puppetlabs-powershell as a dependency
- Tested on Windows Server 2012 R2 and 2016


solarkennedy:
Woot! Thank you.

robklg:
This is great, thank you @iwagnerclgx :)

Issue #399, opened 2018-01-05
Consul should not own its config directory


dekimsey:
https://github.com/solarkennedy/puppet-consul/blob/a8ecd9620591660290bb66ffb4ff331be12f0b5f/manifests/config.pp#L95

Typically services don't own their own config directories. I believe the consul config directory should not be owned by Consul as that makes it writable by the consul service. I think it should be root:consul and 750 by default.

The data directory might also fit into this, it's default is 755, I think it should be 750.

lesinigo:
Same applies to the configuration file, if the config directory shouldn't be world-readable nor consul-writable then `config.json` shouldn't be either.

PR #405, opened 2018-01-30
Set acces rights correctly

Faffnir:
For security concerns it is better to set the access rights so that no other user can execute consul as it can give someone the possibility to join a cluster or issue a cli-command if http_and_cli is active

Also stated in this interesting article: https://www.mauras.ch/securing-consul.html

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -13,12 +13,16 @@ class consul::install {
       $binary_group = 'Administrators'
     }
     default: {
+      if ($consul::manage_group) and ($consul::install_method != 'docker' ) {
+        $binary_group = $consul::group_real
+      } else {
+        $binary_group = 0
+      }
       $binary_name = 'consul'
-      $binary_mode = '0555'
-      $data_dir_mode = '755'
+      $binary_mode = '0750'
+      $data_dir_mode = '0750'
       # 0 instead of root because OS X uses "wheel".
       $binary_owner = 'root'
-      $binary_group = 0
     }
   }
 


solarkennedy:
I'm OK with this PR, but I'm confused about what we are protecting against. Isn't the most important line on :34 and still unchanged?

Making *the binary* not executable doesn't sound like a normal kind of security precaution that is taken in other systems. Anyone can just cp /tmp/ && chmod +x ?

Faffnir:
Absolutely correct about line :34 but I handled that with the already implemented feature of setting the config_mode to 660.

Actually copying it to /tmp shouldn't be possible anymore for anyone or I just miss something here?
What could happen is that anyone uploads a executable onto your server with scp and executes it, but that's outside of this modules scope i think...

What I actually tried to do with this change is to only give the consul::user_real and root ('wheel') the possibility to execute the consul executable which is installed by this module.

lesinigo:
I second that: anyone can copy the binary to `/tmp` if readable or download a new one if not readable, the consul binary in and itself is IMHO equivalent to having access to `curl`.

The actual access control is being able to talk with the cluster or not (network access, [gossip encryption](https://www.consul.io/docs/agent/encryption.html#gossip-encryption), [RPC encryption](https://www.consul.io/docs/agent/encryption.html#rpc-encryption-with-tls), [ACLs](https://www.consul.io/docs/guides/acl.html), and so on...)

lesinigo:
@Faffnir the blog you are linking to actually has a good advice, but the real value does not lie in the consul _binary_ being 0750 rather in the consul _config_ being not-world-readable (to avoid exposing encryption keys or other possible secrets). See #399 for something related to this issue.

Faffnir:
As I stated in the beginning. With this version of puppet-consul it is impossible to copy the consul executable by any other user than root/consul. Of course somebody can download/upload a executable but that's beyond this module or am I mistaken? (you should handle this with other tools?)

Of course the other security features are also very important which @lesinigo listed.

But if I can execute consul, I can also point it to a different config location which makes the whole not readable config location obsolete. I can also try to connect to other hosts and join the cluster which are most likely in the same network and listening on 8301,8302 which could give someone the possibility to read k/v-data. (if you missed one ACL for example)

I would absolutely agree to dismiss this PR if your point is that if ALL the other security measures (which were pointed out by @lesinigo ) are set correctly there is no harm if another user executes consul. Is this the point you are trying to make @lesinigo and @solarkennedy ?

Faffnir:
And it's a good point that the consul config shouldn't be writable by consul @lesinigo ! I missed that... but it can actually be set with the config_mode.

solarkennedy:
I agree with @lesinigo. #399 is probably this module's largest security target.

PR #406, opened 2018-01-31
Support space-separated list in addresses

sw0x2A:
> In Consul 1.0 and later [addresses from config_hash] can be set to a space-separated list of addresses to bind to [...]

 https://www.consul.io/docs/agent/options.html#addresses

Using a space-separated list in addresses for 'http' is currently not supported because it would break config in init scripts and other places. Limiting this to the first one mentioned fixes this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -208,7 +208,7 @@ class consul (
   }
 
   if ($config_hash_real['addresses'] and $config_hash_real['addresses']['http']) {
-    $http_addr = $config_hash_real['addresses']['http']
+    $http_addr = split($config_hash_real['addresses']['http'], ' ')[0]
   } elsif ($config_hash_real['client_addr']) {
     $http_addr = $config_hash_real['client_addr']
   } else {


solarkennedy:
Instead of splitting and using the first entry, I would rather we just send the space-separated thing down to all the places. All the things that *use* http_addr should be able to handle the space-separated notation (right?).

sw0x2A:
I would agree but all the cases where http_addr is used accept only one IP. 

The config_hash is used to generate the config file where it needs the space-separated list.

The http_addr variable is only used to in cases to specify _one_ address to query HTTP API: Init scripts, CLI parameter -http-addr=$http_addr:$port. I am not aware of a way to set multiple host:port in these cases.

sw0x2A:
Let me rephrase my answer. Instead of having all the things that use http_addr be able to handle the space-separated notation, I would introduce another variable to not have the same logic spread across the module and in templates. One with the first IP of the list, the other with the entire list. Since all places where the current http_addr is used are to give a value to the -http-addr CLI parameter, which is a single IP only, I changed the existing variable.

I could have introduced another, new variable with the entire list but it wouldn't be used by now, so I did not.

PR #408, opened 2018-02-16
Add new parameters to configure consul permissions in the binary

cristianjuve:


manifests/install.pp
```diff
@@ -13,12 +13,11 @@
       $binary_group = 'Administrators'
```
solarkennedy: As designed, this PR will not take effect on windows installations.

Can you move these into params so that the too can be set via class params?

PR #412, opened 2018-04-04
added install prefix to dirctory list to avoid puppet error when usin…

kdimiche:
…g default data dir.

```
Error: Cannot create /opt/consul/archives; parent directory /opt/consul does not exist
Error: /Stage[main]/Consul::Install/File[/opt/consul/archives]/ensure: change from absent to directory failed: Cannot create /opt/consul/archives; parent directory /opt/consul does not exist
```

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -30,6 +30,7 @@ class consul::install {
 
       include archive
       file { [
+        $install_prefix,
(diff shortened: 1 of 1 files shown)


solarkennedy:
Can you take a look at the failing tests?

PR #416, opened 2018-05-09
enable acceptance tests on travis

bastelfreak:


bastelfreak:
so I had the feeling that there is a bug on centos-7. The tests prove that. I will work a fix later if I find the root cause.

bastelfreak:
@solarkennedy can you please review and merge this? I pulled the centos tests out into another PR, so we can proceed here.

solarkennedy:
Thanks! These are super tricky to get right.

PR #418, opened 2018-05-14
Change enableTagOverride to enable_tag_override in Consul 1.1.0 and later

hfm:
Consul 1.1.0 has dropped support for enableTagOverride. Use enable_tag_override instead.

see also: https://github.com/hashicorp/consul/blob/v1.1.0/CHANGELOG.md#110-may-11-2018


solarkennedy:
Thanks!

PR #424, opened 2018-06-22
Add Ubuntu 18.04 / 16.04 support

bastelfreak:
This should be merged after https://github.com/solarkennedy/puppet-consul/pull/416


solarkennedy:
Nice. Can you rebase on master?

bastelfreak:
so in general this works, I only need to convince beaker to install puppet5 on ubuntu18.04 and not puppet 4. There is probably an option for that.

bastelfreak:
@solarkennedy can you please review and merge?

solarkennedy:
Thanks!

PR #437, opened 2018-07-29
Fix broken testmatrix due to gem updates

bastelfreak:


bastelfreak:
We need to do some adjustments to the tests, that requires a bit more time.

bastelfreak:
so one thing let to another and yeah.... it got a bit bigger than expected.

bastelfreak:
I have no idea why it fails on puppet 4 :(

solarkennedy:
Green checkmark!

metadata.json
```diff
@@ -46,33 +46,33 @@
...
-        "6.0",
-        "7.0"
+        "5",
+        "6",
+        "7"
       ]
     },
     {
-      "operatingsystem": "Scientific Linux",
+      "operatingsystem": "Scientific",
```
solarkennedy: These `operatingsystem` changes surprised me. Also operatingsystemrelease. This matches what stdlib has, so I buy it.

solarkennedy:
This is a pretty huge deal. Thank you for spending (I'm sure) hours on this. I know how hard it can be to keep things in sync as gems change, *especially* beaker.

This is a reasonably large refactor, do you think I should combine this with a major version release? With the `facts[]` stuff, I'm sure there is a reasonable subset of people who are using this module on old versions of puppet and it "works", but won't work after this (right?).

metadata.json
bastelfreak: I checked facterdb, and it looks like correct name of the operatingsystem is just `Scientific` and not `Scientific Linux`

bastelfreak:
The module currently supports Puppet 4.7.1 and newer, the `facts` hash is available since 4.0. In my opinion a major release isn't required. But we should probably bump the default consul version we install to the latest one and that qualifies as a major release.

solarkennedy:
Got it. Merge when ready!

bastelfreak:
@solarkennedy IMO this is ready. Or do you have some parts that you would like to have changed?

solarkennedy:
Nope!

PR #439, opened 2018-09-11
Update systemd dependency

l-lotz:
Allow newer versions of camptocamp/systemd

PR #440, opened 2018-09-13
Allow puppetlabs-stdlib v5.x

hfm:


shamil:
@bastelfreak I think this PR depends on #439, otherwise tests will fail

PR #444, opened 2018-10-16
Puppet 6 support

l-lotz:
This pull request makes the consul module run under puppet 6. It converts the functions from the legacy puppet 3.x API to the new puppet function API. It also adds unit tests to the travis jobs for puppet6 to ensure future compatibility.


metadata.json
```diff
@@ -11,7 +11,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.7.1 < 6.0.0"
+      "version_requirement": ">= 4.7.1 < 7.0.0"
```
bastelfreak: do the new functions actually work with puppet 4.7.1 or do they require puppet5/4.10?

l-lotz: In the docs it states that the legacy API should only be used for puppet 3.x compatibility, so I assumed that puppet 4.x should be supported. The [docs for 4.9](https://puppet.com/docs/puppet/4.9/functions_ruby_overview.html) mention both APIs so at least 4.9 should be supported (these pages do not exist for 4.8 and 4.7 unfortunately).

bastelfreak:
Thanks. I will mark this as `backwards-incompatible` since the function are new namespaced.

PR #456, opened 2018-12-12
Don't monkey-patch the JSON module

### LINKED ISSUE #452: Broken JSON module
With latest version (5.x) of this module I can't any longer use `JSON` module.

Seems like after #444 the `JSON` module is no longer usable outside the `consul::sorted_json` function.
Seems like overriding  modules with [The modern Ruby functions API](https://puppet.com/docs/puppet/6.0/functions_ruby_overview.html) causing problems.

joshuaspence:
Fixes #452.

### DIFF
--- a/lib/puppet/functions/consul/sorted_json.rb
+++ b/lib/puppet/functions/consul/sorted_json.rb
@@ -1,101 +1,11 @@
 require 'json'
 
-module JSON
-  class << self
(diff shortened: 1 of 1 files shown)


bastelfreak:
@joshuaspence thanks! I will try to test this locally as soon as possible.

PR #458, opened 2018-12-14
Restart only if config.json is chenged

kolaCZek:
Restart only if config.json is changed. Removing service definition executes only service reload.

Signed-off-by: Martin Kolaci <martin.kolaci@lmc.eu>

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -105,6 +105,7 @@ class consul::config(
     group   => $consul::group_real,
     purge   => $purge,
     recurse => $purge,
+    notify  => Class['consul::reload_service'],
   }
   -> file { 'consul config.json':
     ensure  => present,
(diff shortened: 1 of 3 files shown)


solarkennedy:
Can you describe again why we need this change? Seems like a good idea in general to reload if anything in this class changes, yes?

kolaCZek:
It's difference between reload and restart. If you remove service definition -> config file for service is removed (purge on config folder) -> restart is invoked -> node is removed from consul for a moment. This pull request does only reload if service is removed. But still restarts service if config.json is changed.

bastelfreak:
Hi @kolaCZek, can you please take a look at the failed travis jobs? Also please have a look at the used email address in the commit, it isn't associated with your github account.

kolaCZek:
It should be fixed now...

PR #459, opened 2019-01-02
Change allow for spaces in the path, Add extra_options to the win agent

monkey670:
When running with a path containing spaces the run fails. Surrounding the path in quotes and adding & to the front of the exec it now works.

I've also added the extra_options to the consul agent arguments so that additional parameters can be passed to the executable.

### DIFF
--- a/manifests/windows_service.pp
+++ b/manifests/windows_service.pp
@@ -17,7 +17,7 @@ class consul::windows_service {
   $app_dir = regsubst($consul::bin_dir, '\/', '\\', 'G')
   $app_exec = "${app_dir}\\consul.exe"
   $agent_args = regsubst($consul::config_dir, '\/', '\\', 'G')
-  $app_args = "agent -config-dir=${agent_args}"
+  $app_args = "agent -config-dir='${agent_args}' ${consul::extra_options}"
   $app_log_path = "${app_dir}\\logs"
   $app_log_file = 'consul.log'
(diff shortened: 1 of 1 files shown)


bastelfreak:
Can you explain why the `&` is needed?

monkey670:
Once the path is quoted powershell will treat it as just a string. The & is the call operator which will force powershell to execute the string as a command.  For example if I try to run a script located in C:/Program Files (x86)/example.ps1 containing Write-Host "Script Has Run"

Without any quoting this is similar to the error I was getting: 
```
PS C:\Windows\system32> C:/Program Files (x86)/example.ps1
x86 : The term 'x86' is not recognized as the name of a cmdlet, function,
script file, or operable program. Check the spelling of the name, or if a path
was included, verify that the path is correct and try again.
At line:1 char:19
+ C:/Program Files (x86)/example.ps1
+                   ~~~
    + CategoryInfo          : ObjectNotFound: (x86:String) [], CommandNotFound
   Exception
    + FullyQualifiedErrorId : CommandNotFoundException
```

With quoting but no call operator:
```
PS C:\Windows\system32> 'C:/Program Files (x86)/example.ps1'
C:/Program Files (x86)/example.ps1
```

Finally with a call operator
```
PS C:\Windows\system32> & 'C:/Program Files (x86)/example.ps1'
Script Has Run

```

bastelfreak:
Thanks!

PR #460, opened 2019-01-10
Add service_config_hash to customize services

dan-wittenberg:
This will allow you to add new features, such as the new connect options.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -27,11 +27,35 @@
 # [*service_name*]
 #   Name of the service. Defaults to title.
 #
+# [*service_config_hash*]
+#   Use this to populate the basic service params for each of the services
+#
 # [*tags*]
 #   Array of strings.
 #
 # [*token*]
 #   ACL token for interacting with the catalog (must be 'management' type)
+#
+# === Examples
+# @example
+#   ::consul::service { 'my_db':
+#    port                => 3306,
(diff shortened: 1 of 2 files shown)


manifests/service.pp
```diff
@@ -41,6 +65,7 @@
   $id                  = $title,
   $port                = undef,
   $service_name        = $title,
+  $service_config_hash = [],
```
bastelfreak: Could you add a datatype to the parameter? Also in your example it's a hash, here it is an array.

dan-wittenberg: Gotcha, wasn't clear on the datatypes since some had and some don't, and there weren't any in the service manifest.  All should be good now.

bastelfreak: It takes some time to convert all old code to datatypes, but I think it should be enforced for new parameters.

manifests/service.pp
```diff
@@ -41,6 +65,7 @@
   $id                  = $title,
   $port                = undef,
   $service_name        = $title,
+  Hash $service_config_hash = {},
```
bastelfreak: Can you please align the `=` chars and enforce the structure of the Hash?

dan-wittenberg: Alignment done, enforcement delayed per our slack conversation

Issue #461, opened 2019-01-15
Change service resource to file name based on service name rather than ID


jstewart612:
Using ID can create issues when the ID causes the file name to become too long.

IDs have less restrictions in the Consul system.  Names have additional restrictions which ensure the file name will never cross maximum file length if you use it.

jstewart612:
https://github.com/solarkennedy/puppet-consul/pull/462 submitted

PR #462, opened 2019-01-15
$service_name instead of $id for file name

jstewart612:
This is to prevent file names from being too long for the guest OS as per #461


bastelfreak:
@jstewart612 thanks for the PR. Can you take a look at the failing spec tests?

PR #467, opened 2019-02-04
Don't write out meta parameter when unset

jarro2783:
For versions of Consul < 1.1.0, setting `"meta": {}` in the service config doesn't work, because Consul doesn't know about the key. This sets it to `undef` so that there won't be a meta hash with no keys written out when unspecified.


bastelfreak:
thanks!

PR #470, opened 2019-02-13
fix case where multiple http_addr

robmbrooks:
with patch applied...

Notice: /Stage[main]/Consul::Config/File[/etc/init.d/consul]/content: 
--- /etc/init.d/consul	2019-01-30 11:59:11.133195366 -0800
+++ /tmp/puppet-file20190213-127228-1atiisu	2019-02-13 10:57:09.024947000 -0800
@@ -17,7 +17,7 @@
 CONFIG=/etc/consul
 PID_FILE=/var/run/consul/consul.pid
 LOG_FILE=/var/log/consul
-HTTP_ADDR=-http-addr=127.0.0.1 10.1.2.3:8500
+HTTP_ADDR=-http-addr=127.0.0.1:8500
 
 [ -e /etc/sysconfig/consul ] && . /etc/sysconfig/consul

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -235,7 +235,7 @@ class consul (
   if ($config_hash_real['addresses'] and $config_hash_real['addresses']['http']) {
     $http_addr = split($config_hash_real['addresses']['http'], ' ')[0]
   } elsif ($config_hash_real['client_addr']) {
-    $http_addr = $config_hash_real['client_addr']
+    $http_addr = split($config_hash_real['client_addr'], ' ')[0]
   } else {
     $http_addr = $::ipaddress_lo
   }


bastelfreak:
Hi @robmbrooks, aren't multiple ip addresses allowed here? If so, should we instead enforce it in the datatype of the parameter instead of choosing always the first one? Can you also have a look at the used email address in your commits? It isn't associated with your github account.

robmbrooks:
> aren't multiple ip addresses allowed here?

within config.json yes, but used within the init script they don't work

> $ service consul status
> /etc/init.d/consul: line 20: 10.1.2.3:8500: command not found

the following content in the init script is the issue:

> HTTP_ADDR=-http-addr=127.0.0.1 10.1.2.3:8500

which is used here...

> "$CONSUL" info ${HTTP_ADDR}

given both addresses are listening I chose the first, 

I can fix my email address if it is an issue, it just happens I committed from work.

robmbrooks:
edit my commit email and force pushed

robmbrooks:
@bastelfreak can you review when you have a chance

bastelfreak:
@solarkennedy I'm fine with this. Can you take a look as well?

solarkennedy:
I'm fine with it as well.
It is kinda related to 
https://github.com/solarkennedy/puppet-consul/pull/406


In fact, it is kinda a regression. @robmbrooks can you look at the above PR and see if you can solve it in a general way like that, instead of only in the systemd file?

robmbrooks:
@solarkennedy I will adjust the PR to fix all the distro init/upstart scripts. Ironically systemd is the only one that needs no adjustment because systemd does not rely on a scripted status and instead relies on systemd process monitoring.

robmbrooks:
@solarkennedy hopefully this approach is more tasteful, and it copies the approach taken for multiple $config_hash_real['addresses']['http'] values.

solarkennedy:
wfm

PR #436, opened 2018-07-29
implemented service meta parameters

sc0rp10:
Fixed #435

PR #473, opened 2019-03-28
quote all values of tags, meta & node_meta

### LINKED ISSUE #283: Quoted integer (string) to integer is breaking tags

### LINKED ISSUE #468: Can't use integer on meta hash on services
Hey there,

after #466 made it into master we started to play with it.
We quickly found out that passing in a integer as value for a key prevents Consul from reloading.

```puppet
::consul::service { 'foobar':
  checks => [
    {
      tcp      => '127.0.0.1:22',
      interval => '60s'
    }
  ],
  port   => 22,
  meta   => {
      sla => 1
  }
}
```

causes

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Consul::Service[foobar]: parameter 'meta' entry 'sla' expects a String value, got Integer  at /etc/puppetlabs/code/modules/REDACTED/manifests/init.pp:29 on node REDACTED
```

while 

```puppet
::consul::service { 'foobar':
  checks => [
    {
      tcp      => '127.0.0.1:22',
      interval => '60s'
    }
  ],
  port   => 22,
  meta   => {
      sla => '1'
  }
}
```

makes Puppet happy but prevents a Consul reload due to

```
~# consul validate /etc/consul
Config validation failed: Error parsing /etc/consul/service_foobar.json: 1 error(s) decoding:

* 'service.meta[sla]' expected type 'string', got unconvertible type 'float64'

~# cat /etc/consul/service_foobar.json
{
    "service": {
        "checks": [
            {
                "interval": "60s",
                "tcp": "127.0.0.1:22"
            }
        ],
        "enable_tag_override": false,
        "id": "foobar",
        "meta": {
            "sla": 1
        },
        "name": "foobar",
        "port": 22,
        "tags": [
            
        ]
    }
}

```

Might the function `consul::sorted_json`, or the inner called `to_json`, called in the define `::consul::service` cause this?
As I've no clue about these ruby functions so help is needed here.

We're running on Consul 1.3.1 on Ubuntu 16 with Puppet 4.10.14.

Best regards 

jardleex

tmu-sprd:
Values of tags, meta and node_meta needs always to be quoted.

This fixes #318, fixes #468, fixes #283 and closes #436.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -71,7 +71,11 @@ define consul::service(
   Hash $service_config_hash                  = {},
   $tags                                      = [],
   $token                                     = undef,
-  Optional[Hash[String[1], String[1]]] $meta = undef,
+  Optional[Hash[
+    String[1],
+    Variant[
+      String[1], Numeric, Boolean
+  ]]]                                  $meta = undef,
 ) {
 
   include consul
--- a/lib/puppet/functions/consul/sorted_json.rb
+++ b/lib/puppet/functions/consul/sorted_json.rb
@@ -57,17 +57,22 @@ Puppet::Functions.create_function(:'consul::sorted_json') do
   #     }
   #
   def sorted_json(unsorted_hash = {}, pretty = false, indent_len = 4)
+    quoted = false
     # simplify jsonification of standard types
     simple_generate = lambda do |obj|
       case obj
         when NilClass, :undef
           'null'
         when Integer, Float, TrueClass, FalseClass
-          "#{obj}"
+          if quoted then
+            "\"#{obj}\""
+          else
+            "#{obj}"
+          end
         else
           # Should be a string
           # keep string integers unquoted
-          (obj =~ /\A[-]?(0|[1-9]\d*)\z/) ? obj : obj.to_json
+          (obj =~ /\A[-]?(0|[1-9]\d*)\z/ && !quoted) ? obj : obj.to_json
       end
     end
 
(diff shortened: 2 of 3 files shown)


manifests/service.pp
```diff
@@ -71,7 +71,11 @@
   Hash $service_config_hash                  = {},
   $tags                                      = [],
   $token                                     = undef,
-  Optional[Hash[String[1], String[1]]] $meta = undef,
+  Optional[Hash[
```
bastelfreak: thanks for the PR! What do you think about making this not optional and default to an empty hash? Would just make the datatype a bit more readable :thinking: . Please update if you think that's a good idea or just comment here. Afterwards I can merge it.

tmu-sprd: AFAIK, making it non-optional may break older consul versions <1.1.0 (see #467).

I tried to make the datatype more readable by stretching it to multiple lines instead of a long line with the equal sign far right. Unfortunately, there is no real style guide for extensive data type definitions at class parameters.
How about just breaking into two lines?
```
Optional[Hash[String[1], Variant[
    String[1], Numeric, Boolean]]]  $meta = undef,
```

The idea behind the long data type is, to give the user the freedom to choose between using quoted and unquoted values in data yamls, but making sure only working data types can be used.

bastelfreak: okay, let's keep it as it is.

PR #475, opened 2019-04-11
Switch from anchor pattern to contain function

natemccurdy:
The `contain` function is preferred over the anchor pattern when on Puppet 4+.

This also:
* Removes the resource-like declaration of the sub classes which makes the code a bit more portable.
* Accounts for the `$consul::notify_service` variable not existing anymore in `init.pp`


bastelfreak:
thanks!

PR #476, opened 2019-04-11
Lint, style, and parameter data types

natemccurdy:
This PR fixes various style issues that I've had to fix internally and figured I should share them upstream:

* Add parameter data types to all class parameters.
* Use `params.pp` only for calculating default values, not static values.
* Whitespace and indentation cleanup
* Removes redundant relationships between resources.
* Use modern facts and the `facts` hash where possible.
* Use `dig()` instead of checking for sub-key existence in a Hash.


bastelfreak:
@natemccurdy can you please rebase?

natemccurdy:
@bastelfreak Yup! Thanks for reviewing so quickly.

Rebased and pushed 👍

bastelfreak:
Thanks for the work @natemccurdy !

pahoughton-sf:
The change of consul::service port parameter type has broken my use of hiera lookup() for my ports

```
# service.yml

# agate port used in multiple places via lookup('agate::port')
agate::port: 4464

agate::consul_svc:
  agate:
    port: "%{lookup('agate::port')}"
    meta:
      example: value

```
```
# agate.pp

class agate( $port, $consul_svc ) {
  # ...
  create_resources('consul::service', $consul_svc)
```

What are your thoughts on reverting port back to a string value?

natemccurdy:
@pahoughton-sf The `lookup()` function casts everything to a String. I can see that you are defining the port, `4464`, as an Integer. If you switch to the `alias()` function, the data type will be preserved.

```yaml
agate::port: 4464

agate::consul_svc:
  agate:
    port: "%{alias('agate::port')}"
    meta:
      example: value
```
Does that work for you?

If not, I could submit a PR that opens up the type restriction to allow Integers or Strings, but I'd prefer not to. Mainly because the standard type for a port range is generally an Integer: `Integer[0, 65535]`. This is based on the `Stdlib::Port` data type from puppetlabs/stdlib: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/port.pp

For reference, here's a type that allows integers and strings that look like numbers:
`Variant[Integer[0,65535],Pattern[/\d+/]]`

pahoughton-sf:
The alias will work perfect for my use case.  Thanks!

PR #480, opened 2019-05-21
Fixing idempotence issues of new ACL system + improved handling of properties

### LINKED ISSUE #479: New 1.4+ ACL support is not idempotent
The new support for Consul 1.4+ ACL functionality (added in https://github.com/solarkennedy/puppet-consul/pull/474) appears to not be idempotent.

### Expected Behaviour

```
  consul_policy { "my-policy":
    ensure        => 'present',
    description   => "my-policy",
    rules         => [
      {
        'resource'    => 'key_prefix',
        'segment'     => '',
        'disposition' => 'write',
      },
    ]
  }
```

When Puppet is first run, the policy should be created, and a notice should appear such as `Created Consul ACL policy my-policy with ID 315bab4b-ed75-7537-a722-757abca4d16b`.

On further runs, nothing should happen, and nothing should register as changed.

### Actual Behaviour

When Puppet is first run, the policy is created, and a notice appears, such as `Created Consul ACL policy my-policy with ID 315bab4b-ed75-7537-a722-757abca4d16b`.

However, on the first run and all subsequent runs, other notices also appear, such as `/Stage[main]/Consul_policy[my-policy]/id: defined 'id' as '315bab4b-ed75-7537-a722-757abca4d16b'`, which are registered as a change in Puppet.

### Further Information

This non-idempotent behaviour causes Puppet to report that something has changed every time it runs, even though everything on the Consul side is still exactly the same. Once a policy or token has been created, if the module doesn't change anything on subsequent runs, nothing should be reported as changed.

SkydiveMarius:
Fixes #479 and improves handling of some properties (better tracking of modifications).

SecretID of token is not displayed for security reasons. Anyway this property may be only set upon creation, so no modification tracking needed.

Feedback is highly welcome.

### DIFF
--- a/lib/puppet/provider/consul_policy/default.rb
+++ b/lib/puppet/provider/consul_policy/default.rb
@@ -67,6 +67,12 @@ Puppet::Type.type(:consul_policy).provide(
     @client = client
     @rules_encoded = rules_encoded
     @existing_policy = existing_policy
+
+    if existing_policy
+      @property_hash = {
+          :id => existing_policy.id
+      }
+    end
   end
 
   def exists?
--- a/lib/puppet/provider/consul_policy/default.rb
+++ b/lib/puppet/provider/consul_policy/default.rb
@@ -35,7 +35,7 @@ Puppet::Type.type(:consul_policy).provide(
         existing_policy = nil
       end
 
-      resource.provider = new({}, @client, rules_encoded, existing_policy)
+      resource.provider = new({}, @client, rules_encoded, existing_policy, resource)
     end
   end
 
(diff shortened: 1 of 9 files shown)


SkydiveMarius:
Travis checks seem to fail randomly, as Consul seems to be unavailable:
> Connection refused by API after 10 tries

@solarkennedy Is there a way to show the Consul log in case of a failure?

NFarrington:
I can confirm this has fixed my use-case for the issue I reported in #479 🙂

I have spotted one other remaining idempotency issue - marking a policy or token as 'ensure => absent' always causes a notification, such as `Notice: /Stage[main]/Consul_token[my-token]/ensure: removed`, even if the token has already been removed

Thanks for your work on the original feature and this fix, it's helped me a lot!

SkydiveMarius:
@NFarrington Thanks for the feedback!
I just adjusted the creation checks and added an acceptance test.
Could you please verify again?

NFarrington:
All checks out on my end, looks good 👍

solarkennedy:
Thanks @SkydiveMarius for the followup. Nice to catch these type of bugs before we make a release.
Thanks for reporting @NFarrington, just be careful using this from master :)

PR #481, opened 2019-05-31
Remove the id=title variable, make event_name=title

robbyt:
Otherwise, when creating a watch event, nameless events are easily created.

### DIFF
--- a/manifests/watch.pp
+++ b/manifests/watch.pp
@@ -13,7 +13,7 @@
 #   Defaults to 'present'
 #
 # [*event_name*]
-#   Name of an event to watch for.
+#   Name of an event to watch for, defaults to the resource title.
 #
 # [*handler*]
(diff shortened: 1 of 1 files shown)


solarkennedy:
Seems fine. Can you take a look at the failing tests? (they look related to this change)

robbyt:
Thanks for looking at this so quickly @solarkennedy,

The fix for the tests failing isn't great- I had to add another parameter to the defined resource type to keep the previous behavior. The tests assume the config filename is the resource title, not the event_name, and so they were failing because of a different config filename.

Two possible fixes that I can see:
1. Update the test, and potentially change everyone's config filename using this module
2. Use the latest code I pushed that adds a new `config_filename` parameter (which is a bit ugly)

What do you think?

solarkennedy:
I vote option 1.

PR #494, opened 2019-09-24
Add a sleep between tries of consul service reload

cmd-ntrf:
As expressed issue #231, consul reload will fails if it happens while consul is still booting. However, the addition of multiple tries does not solve the issue as all tries can occur in a very short lapse of time.

In this PR, we introduce a sleep between each try, leaving consul up to 20 seconds between the first and the last reload to complete its boot, which should be enough.

### DIFF
--- a/manifests/reload_service.pp
+++ b/manifests/reload_service.pp
@@ -42,6 +42,7 @@ class consul::reload_service {
       command     => $command,
       refreshonly => true,
       tries       => 3,
+      try_sleep   => 10,
     }
   }
 }


bastelfreak:
I never understood why we have this exec resource. Can't we send a notify to the service resource?

solarkennedy:
The original intent was to make it so we could do a lighter weight `reload` instead of a full restart.

solarkennedy:
@cmd-ntrf can you add a trailing comma and rebase on master?

cmd-ntrf:
done.

PR #495, opened 2019-09-26
systemd template: mention it's managed by puppet

bastelfreak:


solarkennedy:
This puppet6 failure on redhat has been failing forever right? Hmm...

Issue #496, opened 2019-09-27
upgrade puppetlabs-stdlib version in dependencies to < 7.0.0?


jay7x:
Current `puppetlabs-stdlib` version is 6.1.0. Is it possible to bump it in dependencies to include this version in range?

bastelfreak:
Hi, thanks for the hint. Can you provide a PR with this change?

PR #497, opened 2019-09-27
Bump upper version range for puppetlabs/stdlib

jay7x:
Current stdlib version is 6.1.0 so it's good idea to bump upper version range to 7.0.0

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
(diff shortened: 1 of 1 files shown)


jay7x:
This closes #496

jay7x:
It'd be great to release v0.5.2 with this change if possible

bastelfreak:
@jay7x can you please also bump the dependency for the archive module? It currently conflicts with stdlib 6.x.

jay7x:
I'll close this one and do proper PR instead

PR #500, opened 2019-09-28
remove failing tests on centos6 + puppet 6

solarkennedy:
Our test coverage is pretty comprehensive, but this centos6 build has been failing for about 2 months.

Bisecting it, here is the last green build:
https://travis-ci.org/solarkennedy/puppet-consul/jobs/551129989
And then the next red one:
https://travis-ci.org/solarkennedy/puppet-consul/jobs/562904128

Both use consul 1.2.3. 

The one that fails has puppet agent 6.7. The Green build has puppet agent 6.5.

Here is a diff
https://pastebin.com/ntQXi0d9

I bet there is something "real" here, but I don't feel like spending time debugging it. Till then I think the broken test isn't providing any value. RFC?


bastelfreak:
@solarkennedy should we also remove centos6 from the metadata.json? Also there was a recent change to the github-changelog gem and it seems it broke our whole testmatrix :(

PR #504, opened 2019-10-09
added support for reload over https

BCarette:
# Why?
When adding a service, or performing other change which requires consul to reload, consul is reloaded with the command:
```
consul reload -http-addr=127.0.0.1:8500 (with address and port configurable)
```
However this assumes that consul allows connections via http, without encryption or identity verification. 

For sensitive productions scenarios however, consul should only allow mutual tls connections. In this case the reloading will fail, as it tries to connect over http without valid certificate.

This merge request will allow the reload to happen over TLS, with the correct certificates.

# How does it work?
If the http port is disabled (equal to -1) and `verify_incoming` is set to true in the consul configuration, the reload will now happen automatically over TLS with the correct certificates.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -277,7 +277,7 @@ class consul (
   if dig($config_hash_real,'verify_incoming') {
     $verify_incoming = $config_hash_real['verify_incoming']
   } else {
-    $verify_incoming = False
+    $verify_incoming = false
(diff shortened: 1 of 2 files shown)


bastelfreak:
Thanks for the PR! I restarted the failed travis jobs.

PR #507, opened 2019-10-29
ensure latest is often unwanted

attachmentgenie:


bastelfreak:
@attachmentgenie could you explain a bit why you did this change? Is the Pull Request title actually correct?

attachmentgenie:
@bastelfreak the title is indeed the exact oposite of what this PR is trying to achieve. The reason for this change i 2 fold. The url version has a hardcoded version, where the package ensures latest. This brings the more inline. Second is that consul clusters dont like to have version deltas. The cluster will most likely survive but there is a higher chance of race conditions.

solarkennedy:
My main reasoning for having latest instead of installed at work is to ensure a consistent version, instead of having a package installed to be whatever was latest at the time of install.

Is there anything preventing you from configuring this module to meet your requirements?

attachmentgenie:
> Is there anything preventing you from configuring this module to meet your requirements?

no, everthing is nicely parameterized for me to control the version from a profile. However to me the  'latest' default is not a sane default, hence the PR. But if you feel different feel free to reject this PR, i wont be offended ;-)

PR #512, opened 2019-11-20
systemd: Start consul after network is really up

bastelfreak:
This change is based on systemd documentation from:
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
The network-online.target is reached after an actual ip configuration is up. This is required if consul should bind to a specific ip address.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 /spec/fixtures/modules/
 /tmp/
 /vendor/
+/.vendor/
 /convert_report.txt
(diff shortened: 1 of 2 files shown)


amiryal:
It is confusing, but `network-online.target` does not fully replace `network.target`, as it serves a different function and comes in addition. When you get a chance, please re-add `After=network.target`, so that Consul has a chance to leave the cluster cleanly on host shutdown, before the network goes away. Thanks!

PR #514, opened 2019-11-20
Move consul unit file to /usr/lib/systemd/system, as we install consul like it was packaged

KrzysztofHajdamowicz:
`/etc/systemd/system` is a place for hand-crafted unit files and overrides units from `/usr/lib/systemd/system`.

Creating overrides or drop-ins is impossible, because puppet overwrites such file.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -54,6 +54,7 @@ class consul::config (
       'systemd': {
         systemd::unit_file { 'consul.service':
           content => template('consul/consul.systemd.erb'),
+          path    => '/usr/lib/systemd/system',
           notify  => $notify_service,
         }
       }


bastelfreak:
Hi @KrzysztofHajdamowicz, using `/usr/lib/systemd/system` is bad idea in my opinion as it should only be used for unit files provided by the package manager. As far as I know you can still use drop-ins if the unit is in `/etc/systemd/system`?

KrzysztofHajdamowicz:
> Hi @KrzysztofHajdamowicz, using `/usr/lib/systemd/system` is bad idea in my opinion as it should only be used for unit files provided by the package manager. As far as I know you can still use drop-ins if the unit is in `/etc/systemd/system`?

Unfortunately, you cant (at least) mask and unmask such service, as systemd creates symlink /dev/null to /etc/systemd/system/$unit_file.

bastelfreak:
But why would you want that if you used puppet to explicitly install the unit file?

solarkennedy:
I'm going to close this due to inactivity. I agree with @bastelfreak that /etc/ is the right place for a puppet-managed service file.

PR #516, opened 2019-12-27
fix(reload_service): remove typo in reload_options

bmx0r:
Just to fix this small typo :)


bmx0r:
thanks for approving :)
i don't know how to solve the failing test thought...

Issue #517, opened 2020-01-14
info required - Apply ACL on https consul


wilfriedroset:
Hello

I've successfully managed to setup and configure a consul cluster.
I'm now trying to implement ACL management using puppet. 
When I try to do anything related to them I end up with the following message:

```
consul1 ~ # puppet agent -vt --environment setup_consul  --noop
Warning: Setting 'pluginsync' is deprecated.
   (location: /usr/lib/ruby/vendor_ruby/puppet/defaults.rb:1879:in `block in <module:Puppet>')
Info: Using configured environment 'setup_consul'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Applying configuration version '1578996743'
Warning: Cannot retrieve ACL token list: Got negative API response (Code: 400, Response: Client sent an HTTP request to an HTTPS server.
)
Notice: /Stage[main]/Consul/Consul_policy[dns-request-policy]/ensure: current_value 'absent', should be 'present' (noop)
Notice: Class[Consul]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 1 event
Notice: Applied catalog in 8.67 seconds
```

All of my servers are running with only the https open on 8501 port and self-signed certificates (debian's snake-oil).
I got the same error message when I issue a http request to a https server:

```
consul1 ~ # curl -k http://[redacted]:8501/v1/kv
Client sent an HTTP request to an HTTPS server.
```

Here is my hieradata (some parts are redacted)
```yaml
consul::bin_dir: /usr/bin/
consul::config_dir: /etc/consul.d
consul::pretty_config: true
consul::install_method: package
consul::package_ensure: present
consul::service_enable: true
consul::service_ensure: running
consul::extra_groups:
  - ssl-cert
consul::acl_api_hostname: "%{scope('::ipaddress_eth0')}"
consul::acl_api_protocol: https
consul::acl_api_port: 8501
consul::config_hash:
  encrypt: >
    [redacted]
  bootstrap_expect: 3
  log_level: DEBUG
  bind_addr: "%{scope('::ipaddress_eth0')}"
  client_addr: "0.0.0.0"
  datacenter: "%{scope('::zone')}"
  node_name: "%{scope('::hostname')}"
  disable_update_check: true
  disable_anonymous_signature: true
  server: true
  ui: true
  retry_join:
    - [redacted]
    - [redacted]
    - [redacted]
[...]

PR #519, opened 2020-01-21
SSL support and allow passing of a custom CA

lukebigum:
These patches address #517.

The providers now turn on SSL support in Ruby's Net::HTTP if the Consul API is specified as 'https'.  It also allows setting a custom CA file, for those installations where the CA used for Consul communication is not in the system trust store.

A provisional PR, for peer review and to check my Ruby code.  I haven't added any HTTPS related tests.  My rspec skills for types and providers is sub-par, but I could give some direction of what tests might be useful.  Stubbing every single provider request as HTTPS seemed like a waste of clock cycles.


lukebigum:
@bastelfreak I've fixed the silly things I broke, and added a test per provider to confirm HTTPS works.  I'd appreciate your thoughts on the coverage.

I've also fixed what appears to be a problem with opensuse 13 support - it's not in FacterDB any more? I've monkey patched this by bumping support of opensuse to v15 in `metadata.json`.  That may or may not be the right thing to do.

Last, I'm unsure how to fix the final failure from TravisCI.  It appears to be a problem with the test run talking to the local Docker instance.  The same test failure occurred in the last run of `master`, so I don't think it's something I caused

PR #522, opened 2020-04-08
Add validate_cmd to the config.json

b3n4kh:
Make use of the [Puppet validete_cmd](https://puppet.com/docs/puppet/latest/types/file.html#file-attribute-validate_cmd). I think it makes most sense to valide the whole folder if purge_config_dir is false and some additional configs exist.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -130,6 +130,7 @@ class consul::config (
     group   => $consul::group_real,
     mode    => $consul::config_mode,
     content => consul::sorted_json($config_hash, $consul::pretty_config, $consul::pretty_config_indent),
+    validate_cmd => "${consul::bin_dir}/${consul::binary_name} validate %",
   }
 
 }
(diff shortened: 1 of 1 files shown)


solarkennedy:
Won't this be a bit expensive to run, because it is running the same command on a whole folder multiple times?

Also does this really make sense, I think it really needs the `%` in place for validate_command to do it's thing (validating *before* it deposits the file)

b3n4kh:
I actually [tried](https://github.com/solarkennedy/puppet-consul/pull/522/commits/d0c35de5476987d92083ab6149cbf199538a6220) with the `%` . Unfortunatly only validating this config file does not give you any feedback wheater consul will start or not. Unfortunatly the only way to validate the config in a meaningful way I found was to check the folder. I could add a flag to make this operation optional. (But I fear this will end in an ugly if around the file resource)

solarkennedy:
But without `%`, what is the validate command even doing? It is just validating the *existing* configs before putting down the new config, leaving a kind of timebomb.

b3n4kh:
I get it, it's only evaluated at the second run. I think validate_cmd just doesn't make sense here. Only checking the file config.json would lead to some strange behaviour if you add an additional static configuration to consul.d, so I wouldn't sugest that either. Or do you think adding the single check with (with `%`) is worth it? Otherwise, I think this PR can be closed.

solarkennedy:
If you could figure out a way to make the validate_cmd actually check a file before it gets in place, then I think it would be worth it. Otherwise I vote close.

PR #527, opened 2020-04-24
Make stringification a parameter of the sorted_json helper functions

rtkennedy:
Move the 'quoted'ness of child values to be a function parameter instead of a
local variable to avoid an issue where keys processed after a
node_meta|meta|tags key were inadvertently quoted. This should allow it to
track down the recursion as hopefully was intended.

The escape hatch for SRV weights should no longer be needed.

The base simple_generate() function has been simplified to jsonify
either a string interpolation of int, float, and boolean or the actual value
depending on the quoting request of the parent.

### DIFF
--- a/lib/puppet/functions/consul/sorted_json.rb
+++ b/lib/puppet/functions/consul/sorted_json.rb
@@ -57,18 +57,13 @@ Puppet::Functions.create_function(:'consul::sorted_json') do
   #     }
   #
   def sorted_json(unsorted_hash = {}, pretty = false, indent_len = 4)
-    quoted = false
     # simplify jsonification of standard types
-    simple_generate = lambda do |obj|
+    simple_generate = lambda do |obj, quoted|
       case obj
         when NilClass, :undef
           'null'
         when Integer, Float, TrueClass, FalseClass
-          if quoted then
-            "\"#{obj}\""
-          else
-            "#{obj}"
-          end
+          quoted ? "#{obj}".to_json : obj.to_json
         else
           # Should be a string
           # keep string integers unquoted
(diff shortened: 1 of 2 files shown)


solarkennedy:
Ready for rebase/merge.

rtkennedy:
Still not working on the CentOS 6 consul 1.5.0 test, and not replicable using beaker locally. I have a feeling that the service start on CentOS 6 is just a hair too slow on the Travis build hosts for the policy api calls that immediately follow it in the test manifest. I'll see if there's a way to split the puppet run with a pause locally just to be sure that's it, or if there's a way to make the consul start hold until the service is actually ready for calls.

rtkennedy:
Triggering a rebuild with a close/open.

rtkennedy:
FWIW, the config generated is
`{
    "acl": {
        "default_policy": "allow",
        "down_policy": "extend-cache",
        "enabled": true,
        "tokens": {
            "master": "222bf65c-2477-4003-8f8e-842a4b394d8d"
        }
    },
    "bootstrap": true,
    "bootstrap_expect": 1,
    "client_addr": "0.0.0.0",
    "data_dir": "/opt/consul",
    "datacenter": "east-aws",
    "leave_on_terminate": true,
    "log_level": "INFO",
    "node_name": "foobar",
    "primary_datacenter": "east-aws",
    "rejoin_after_leave": true,
    "server": true,
    "start_join": [
        "127.0.0.1"
    ]
}`
Which looks OK? The server refuses to start and doesn't put anything in the log. I'll add some more debug manifest attempts to see why it's broken but only on CentOS 6.

rtkennedy:
consul validate passes without issue. When applying the manifest 3 times in succession without catching failures it works fine on the first pass. So confused.

bastelfreak:
@rtkennedy can you rebase this against master?

rtkennedy:
Ugh, sorry about the merge of master instead of a rebase. I can redo the branch from scratch if the checks pass and you want it clean.

rtkennedy:
I'll also need to revert the changes to class_spec.rb that were used for debugging.

bastelfreak:
I'm far away from being able to review this properly, but from what I can tell this looks good. I'm wondering why it current'ly doesn't always fail.

rtkennedy:
Closing to clean up the branch and remove debugging.

rtkennedy:
I forgot that recreated branches are impossible to link back to the same PR. The new PR is https://github.com/solarkennedy/puppet-consul/pull/547.

PR #528, opened 2020-04-27
Use a filter instead of delete_undef_values()

rtkennedy:
Use filter instead of delete_undef_values() to keep puppet 6 from
generating json files with 'null' as the value for undefined keys.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -84,7 +84,7 @@ define consul::check (
   }
 
   $check_hash = {
-    check => delete_undef_values($basic_hash),
+    check => $basic_hash.filter |$key, $val| { $val =~ NotUndef },
   }
 
   consul::validate_checks($check_hash[check])
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -105,7 +105,7 @@ define consul::service (
   $basic_hash = $default_config_hash + $service_config_hash
 
   $service_hash = {
-    service => delete_undef_values($basic_hash),
+    service => $basic_hash.filter |$key, $val| { $val =~ NotUndef },
   }
 
   $escaped_id = regsubst($id,'\/','_','G')
--- a/manifests/watch.pp
+++ b/manifests/watch.pp
@@ -136,8 +136,10 @@ define consul::watch (
     }
   }
 
+  $merged_hash = merge($basic_hash, $type_hash)
+
   $watch_hash = {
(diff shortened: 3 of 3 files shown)


solarkennedy:
There are a couple of "undef" related tests failing, can you look at those?

rtkennedy:
It looks like the consul 1.5.0 service failed to start for just CentOS 6 x86_64 and then caused the API calls to fail and get nil values. I'll enable Travis CI for my fork and try to see what's going on, unless there's a better way to run these checks locally to view what json it's producing and what error consul got when starting up.

rtkennedy:
Running beaker locally doesn't fail the tests in the centos 6 container, I'll need to replicate the Ubuntu 14.04 platform that Travis is using.

solarkennedy:
*some* of the tests have been broken for a long time. I saw the undef there so I thought it might be related, but don't try too hard. The same tests look failing on master:
https://travis-ci.org/github/solarkennedy/puppet-consul

Up for suggestions? We could just junk the one failing beaker test. It would be nice to have the unit tests working though.

rtkennedy:
The first three CHECK=test ones seem to be failing because openSUSE 13 isn't populating facts from facterdb? I don't have a ton of experience with these testing frameworks, so that's just a guess. Locally "facterdb 'operatingsystem=OpenSuSE and operatingsystemmajrelease=13 and architecture=x86_64'" returns facts after falling back to facter 2.2.0-2.4.0, but if the Xenial Travis image is forcing a facter version it would be empty.

solarkennedy:
Screw it, can you remove the opensuse tests to get a green build?

rtkennedy:
Here's to hoping I understood how everything worked: https://github.com/solarkennedy/puppet-consul/pull/529. It still won't fix the broken CentOS 6 test for this original PR, but maybe it will fix the first three.

solarkennedy:
It worked :). Ready for rebase/merge.

rtkennedy:
Welp, I have no idea why it's passing now. It was still failing the CentOS6 test, I added debug stuff to dump log files, it passed, I force reset back to before the debugging, and it still passes. The issue appeared to be the 1.5.0 server not electing itself leader fast enough for the 10 tries of API calls with increasing timeouts betwen each try, so 60 total seconds. Maybe. I never got a failure to coincide with a dump of the log files.

PR #529, opened 2020-05-06
Removing OpenSuSE from metadata.json to stop triggering broken tests

rtkennedy:
OpenSuSE 13 only has facterdb entries for facter 2.2-2.4. The
CHECK=test TravisCI tests keep failing due to lack of facter data
for this platform.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -80,13 +80,6 @@
       ]
     },
     {
-      "operatingsystem": "OpenSuSE",
-      "operatingsystemrelease": [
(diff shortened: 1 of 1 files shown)


rtkennedy:
Looks like I guessed right.

solarkennedy:
!!!

Thanks for going the extra mile and "tuning the exact right bolt"

I thought for sure this would be some rubygems thing or a facter bug or who knows! (well, you knew!)

PR #535, opened 2020-05-21
Allow changing the configuration directory and files owner

thias:
This is because it's good practice to keep the service from potentially overwriting its own configuration through some security flaw, meaning having the owner 'root', the group 'consul' and files mode 0640 and directories mode 0750 makes the most sense to keep everyting secure and private.

Note that this works for me because I deploy consul using an rpm where I have set the config directory to have mode 0750, but this is not something this Puppet module will be enforcing. Should a new `config_dir_mode` also be added, defaulting to `undef` to keep backwards compatibility?

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -115,7 +115,7 @@ class consul::config (
 
   file { $consul::config_dir:
     ensure  => 'directory',
-    owner   => $consul::user_real,
+    owner   => $consul::config_owner_real,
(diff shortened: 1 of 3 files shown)


thias:
@solarkennedy : Any feedback on this change? It's fairly straightforward, passes checks and (to me at least) makes a lot of sense security-wise.

solarkennedy:
I think this is fine

Issue #526, opened 2020-04-23
Service port gets quoted with a puppet 6.10 catalog server and consul rejects it


rtkennedy:
With this service configuration:

```
  consul::service { 'node_exporter':
    address => $::hostname,
    checks  => [
      {
        'http'     => 'http://localhost:9100',
        'interval' => '5s',
      }
    ],
    port    => 9100,
    tags    => $consul_tags;
  }
```

consul module version 6.0.0, and a puppet 6.10 catalog server I get a service json file with a quoted port number:

```
{ "service": {"address":"cadws002","checks":[{"http":"http://localhost:9100","interval":"5s"}],"enable_tag_override":false,"id":"node_exporter","meta":null,"name":"node_exporter",**"port":"9100"**,"tags":["node_exporter","metrics","eng"],"token":null}}
```

which results in a consul error: `* 'service.port' expected type 'int', got unconvertible type 'string'`

The sorted_json() function doesn't seem to do this when the same code is run through a puppet 5.5 catalog server.

rtkennedy:
Additionally, the major version of the client doesn't matter. A 5.5 client pointed at a 6.10 catalog server will get the stringified integer, and a 6.10 client pointed at a 5.5 catalog service will get a bare integer.
Removing 'Integer' from 'when Integer, Float, TrueClass, FalseClass' in sorted_json allows it to fall through and not get quoted - though I'm still not sure why 5.5 wasn't hitting the same code path unless it was stringifying it on the way in and having it end up in the else clause.

rtkennedy:
It's looking more like code to turn on the quoted boolean for everything inside node_meta|meta|tags hashes in the 2 sorted*generate functions is instead turning it on for everything that happens after it in the loop.

helenes-r7:
I'm running into this issue as well

bastelfreak:
Hi people. I need to say that I'm unable to reproduce this. I'm running version 6.1.0 of this module with puppetserver 6.12.1 and puppet agent 6.17.0. example consul code:

```puppet
    consul::service { 'node_exporter':
      checks  => [
        {
          name     => 'node_exporter health check',
          http     => 'http://127.0.0.1:9100',
          interval => '10s',
          timeout  => '1s'
        },
      ],
      port    => 9100,
      address => $consul_address,
      tags    => ['node_exporter'],
    }
```

results in:

```json
{
    "service": {
        "address": "***",
        "checks": [
            {
[...]

PR #539, opened 2020-06-12
Service port gets quoted with a puppet 6.10

jose-bashton:
Apparently, this is reported issue where one collaborator suggested this change to keep the service check port value unquoted as it's an integer type (not a string).
https://github.com/solarkennedy/puppet-consul/issues/526   

```
consul[12894]: * 'service.port' expected type 'int', got unconvertible type 'string'
```

### DIFF
--- a/lib/puppet/functions/consul/sorted_json.rb
+++ b/lib/puppet/functions/consul/sorted_json.rb
@@ -63,7 +63,7 @@ Puppet::Functions.create_function(:'consul::sorted_json') do
       case obj
         when NilClass, :undef
           'null'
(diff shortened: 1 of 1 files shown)


rtkennedy:
This was the simple fix I initially proposed, but it's not the complete solution because the recursion causes nested things to be stringified even if you don't want them to be. I wrote up https://github.com/solarkennedy/puppet-consul/pull/527 but was having trouble reproducing a failure in TravisCI just for the RH6 platform, even though it passed beaker tests on a local setup without issue.

PR #541, opened 2020-07-15
add new optional 'datacenters' parameter to consul_policy type

miklinux:
This PR is to add a new 'datacenters' property to consul_policy type thus allowing to create policies that are valid on specific datacenters only.


solarkennedy:
Can you take a look at the failing tests, which have mocks that are not ready for this new param

miklinux:
Unfortunately I'm not very familiar with unit testing.. I don't think to be of any help here, sorry :confused:

PR #547, opened 2020-08-19
Make stringification a parameter of the sorted_json helper functions

### LINKED ISSUE #526: Service port gets quoted with a puppet 6.10 catalog server and consul rejects it

rtkennedy:
Move the 'quoted'ness of child values to be a function parameter instead of a local variable to avoid an issue where keys processed after a node_meta|meta|tags key were inadvertently quoted. This should allow it to track down the recursion as hopefully was intended.

The escape hatch for SRV weights should no longer be needed.

The base simple_generate() function has been simplified to jsonify either a string interpolation of int, float, and boolean or the actual value depending on the quoting request of the parent.

Fixes #526

### DIFF
--- a/lib/puppet/functions/consul/sorted_json.rb
+++ b/lib/puppet/functions/consul/sorted_json.rb
@@ -57,18 +57,13 @@ Puppet::Functions.create_function(:'consul::sorted_json') do
   #     }
   #
   def sorted_json(unsorted_hash = {}, pretty = false, indent_len = 4)
-    quoted = false
     # simplify jsonification of standard types
-    simple_generate = lambda do |obj|
+    simple_generate = lambda do |obj, quoted|
       case obj
         when NilClass, :undef
           'null'
         when Integer, Float, TrueClass, FalseClass
-          if quoted then
-            "\"#{obj}\""
-          else
-            "#{obj}"
-          end
+          quoted ? "#{obj}".to_json : obj.to_json
         else
           # Should be a string
           # keep string integers unquoted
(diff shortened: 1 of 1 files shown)


bastelfreak:
@rtkennedy can you please rebase?

rtkennedy:
Rebased and added in the change to quote args' children per https://github.com/solarkennedy/puppet-consul/pull/548.

bastelfreak:
Hi,
it looks like you merged master into your branch instead of rebasing. Can you please do an actual rebase to have a clean git history.

rtkennedy:
I'll try, but I thought force pushes caused the PR to die.

bastelfreak:
in general it won't close the PR. there are edge cases, for example force pushing a branch that is equal to master. That's the only situation I'm aware of that will close the PR (and deleting the branch).

rtkennedy:
Hopefully that fixed it up, absent the commit message being less interesting than last time. I can amend it with the original text if that's useful.

solarkennedy:
Thanks @rtkennedy ! 

Is there any test case we can add to actually prove this fixes something?

rtkennedy:
I amended the commit message back to the original in case someone is looking through the logs outside of github. As for test cases, making a consul::service with tags and a port triggered the initial issue https://github.com/solarkennedy/puppet-consul/issues/526. But a better test might be a service with all the things, and running a 'consul validate' on the resulting config.

solarkennedy:
So uh, should I click the green button? :)

rtkennedy:
I think so? If there's something else for me to do let me know.

Issue #533, opened 2020-05-20
Home directory not created for consul user


joshberry911:
When the module creates the consul user it sets the home directory to /home/consul but then does not create this directory.

When I then try to run some maintenance tasks as the consul user in cron they fail

{code}
May 20 11:53:01 consul-001 CROND[19775]: (CRON) ERROR chdir failed (/home/consul): No such file or directory
{code}

I realise that I can just create the directory or create the user myself, but could either of these solutions be implemented in the module?

- **managehome** be set to true (optionally I guess, for backwards compatibility)
-  The home directory be set to the consul config directory

solarkennedy:
I would accept a PR to enable managehome, I think that is fine.

PR #550, opened 2020-08-21
consul user: set correct home

bastelfreak:
Fixes https://github.com/solarkennedy/puppet-consul/issues/533


genebean:
Based on #559 I have added the backwards-incompatible label  here so the change log will be updated to reflect that this is a breaking change for existing users. A workaround is included in #575

PR #551, opened 2020-09-10
Add 'consul leave' as ExecStop-parameter for systemd-unit

BaileyRud:


solarkennedy:
Can you look at where it was previously decided that this was not desirable in #85 and confirm if anything has changed?

PR #552, opened 2020-10-07
notify systemd when consul daemon is really started

ymartin-ovh:
Have the following issue on my environment:

puppet runs:
```
# Notice: /Stage[main]/Consul::Run_service/Service[consul]: Triggered 'refresh' from 1 event
# Warning: Cannot retrieve ACL token list: Failed to open TCP connection to localhost:8500 (Address family not supported by protocol - socket(2) for "localhost" port 8500)
```

I suspect that systemd returns to fast and assume that daemon is ready but it is not the case.
Since consul can notify systemd, I change type from simple (default when type is absent) to notify.

External links:
https://www.hashicorp.com/resources/systemd-the-good-parts

### DIFF
--- a/templates/consul.systemd.erb
+++ b/templates/consul.systemd.erb
@@ -5,6 +5,12 @@ After=network-online.target
 Wants=network-online.target
 
 [Service]
+<% if scope.lookupvar('consul::config::config_hash')['retry_join'].nil? or \
+      scope.lookupvar('consul::config::config_hash')['retry_join'].length < 2 -%>
+Type=exec
+<% else -%>
+Type=notify
+<% end -%>
 <% if @allow_binding_to_root_ports == true -%>
 AmbientCapabilities=CAP_NET_BIND_SERVICE
 <% end -%>


solarkennedy:
You are right, and it is set like this in an "upstream" version of this unit file:
https://learn.hashicorp.com/tutorials/consul/deployment-guide#configure-systemd

That's cool!

ymartin-ovh:
Thanks for the link. It seems, we need to keep type=exec for standalone deployment.

chrisboulton:
I was about to go implement support for `Type=notify` in this same module before I came across this change and realised we were just using an out of date version.

Our environment relies on auto-discovery support in Consul, which means our configuration for `retry_join` only has one entry:

```json
  "retry_join": [
    "provider=gce tag_value=.. zone_pattern=.."
  ],
```

I'm wondering if it makes sense to submit a PR that changes this to `Type=notify` always, and what the downside of that change would be? (alternatively maybe we move this into a configurable param and remove the determining logic?)

PR #554, opened 2020-10-08
allowed insecure connection during archive download

alexhill050392:
This resolves the issue with corporate proxies which have issues with the certificate.


solarkennedy:
You need to add your proxy's cert to your trusted CAs.

PR #558, opened 2021-01-11
Add config_name parameter to define the name of the consul config

bogdankatishev:
Add config_name parameter to give users the flexibility to name their config files in the name they want.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -123,9 +123,9 @@ class consul::config (
     recurse => $purge,
   }
 
-  file { 'consul config.json':
+  file { 'consul config':
     ensure  => present,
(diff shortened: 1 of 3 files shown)


solarkennedy:
Seems fine, but can you tell me more about your use-case? Why not just have it be config.json?

bogdankatishev:
We will be using the upstream hashicorp consul package, which includes a Consul service file, which has this line: ConditionFileNotEmpty=/etc/consul.d/consul.hcl. So we will change the Consul config name to: consul.hcl.

And because hcl is JSON-compatible, the way of working will be the same.

solarkennedy:
Hmm.

PR #560, opened 2021-01-14
adding the option to setup the upstream HashiCorp repository

attachmentgenie:
HashiCorp now provides (some) upstream packages, this PR implements the managing of the repo

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -65,6 +65,9 @@ class consul::install {
       }
     }
     'package': {
+      if $consul::manage_repo{
+        include hashi_stack::repo
+      }
       package { $consul::package_name:
         ensure => $consul::package_ensure,
         notify => $do_notify_service,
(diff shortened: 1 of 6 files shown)


solarkennedy:
I don't think we should default this to be true

attachmentgenie:
@solarkennedy not a problem, i pushed the changes requested

manifests/install.pp
```diff
@@ -65,6 +65,9 @@
       }
     }
     'package': {
+      if $consul::manage_repo{
+        include hashi_stack::repo
```
bastelfreak, line 69: Should we also add this?
```puppet
Class['hashi_stack::repo'] -> Package[ $consul::package_name]
```

attachmentgenie: makes sense, just pushed the update

bastelfreak (approved):

PR #564, opened 2021-05-10
OS and Puppet versions update

genebean:
This updates the list of supported OSs to those that are not end-of-life and updates the versions of Puppet to those that are currently supported. It also updates dependencies to module versions which include support for Puppet 7.

For reference, a CI overhaul that switches to GitHub Actions will be coming in a follow up PR - thus the minimal modifications to the travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,63 +4,21 @@ before_install: rm Gemfile.lock || true
 sudo: false
 cache: bundler
 script: bundle exec rake $CHECK
-matrix:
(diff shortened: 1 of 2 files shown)


genebean:
@bastelfreak & @solarkennedy I think this one is ready to go now. It is a stepping stone in the updates I started earlier as the other PR was getting too big and needed to be split up into multiple PRs

solarkennedy:
Ship all the things!

PR #573, opened 2021-05-11
List Debian 10 as supported

genebean:
This module has long worked on Debian but it had be omitted from the list of supported OSs. This adds it to the list and also sorts the list alphabetically.


genebean:
I think the CI failure is a transient fwiw.

PR #575, opened 2021-05-17
Make home directory location setting optional

### LINKED ISSUE #559: New Home Attribute on consul user resource breaks our puppet runs.
This merge https://github.com/solarkennedy/puppet-consul/pull/550 breaks our Puppet runs on our current Consul Agents.

```
Error: Could not set home on user[consul]: Execution of '/usr/sbin/usermod -d /opt/consul/data consul' returned 8: usermod: user consul is currently used by process 894
Error: /Stage[main]/Consul::Install/User[consul]/home: change from '/home/consul' to '/opt/consul/data' failed: Could not set home on user[consul]: Execution of '/usr/sbin/usermod -d /opt/consul/data consul' returned 8: usermod: user consul is currently used by process 894
```

As far as I can see, many popular Puppet Modules avoid the User resource, or when they use them, they do not set the home attribute.

How could this be resolved cleanly?

genebean:
Fixes #559

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -117,6 +117,12 @@
 # [*manage_user*]
 #   Whether to create/manage the user that should own consul's configuration files.
 #
+# [*manage_user_home_location*]
+#   Whether to explicitly set the location of the consul user's home directory when this modules
+#   manages the creation of the user (aka `manage_user = true`). If the consul user already exists
+#   and this is enabled, puppet tries to change the consul user's home to the new location. This
+#   will cause the puppet run to fail if the consul service is currently running.
+#
 # [*manage_data_dir*]
 #   Whether to manage the consul storage data directory.
 #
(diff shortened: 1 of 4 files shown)


genebean:
@bastelfreak I am not sure I picked the right label here - feel free to change if I goofed

bastelfreak (approved):
I did some chatting with @genebean on IRC and I think  this is the best approach to move forward. It worka for new and existing environments and in my opinion that is important.

genebean:
@bastelfreak how do you feel about merging this and cutting a release?

manifests/init.pp
```diff
@@ -117,6 +117,12 @@
 # [*manage_user*]
 #   Whether to create/manage the user that should own consul's configuration files.
 #
+# [*manage_user_home_location*]
+#   Whether to explicitly set the location of teh consul user's home directory when this modules
```
solarkennedy, line 121: `teh`

manifests/init.pp
```diff
@@ -117,6 +117,12 @@
...
+#   manages the creation of the user. If the consul user already exists and this is enabled, puppet
```
solarkennedy, line 122: Can you also note that this setting only takes effect if `manage_user` is true ?

solarkennedy (approved):

manifests/init.pp
genebean: fixed

manifests/init.pp
genebean: How does that look @solarkennedy ?

PR #577, opened 2021-06-03
enable ssl when protocol is https

SimonPe:
This is a minimal implementation of #519 and fix for #517
This just enables actual ssl use, without adding any extra options for CA's and such.


solarkennedy:
Looks... pretty easy!
Let's make sure the tests pass then I'll merge.

Issue #407, opened 2018-02-07
Add Consul's autocompletion for shell


jardleex:
Hi,

I just stumbled over the fact that [Consul supports autocompletion](https://www.consul.io/docs/commands/index.html#autocompletion). 
It would be nice if this module would support that.

Issuing the command documented with a Puppet exec ain't worked for us as it seem to require a interactive shell. To bypass this we solved it this way:

```
      file_line { 'enabe_consul_autocompletion':
        path  => '/root/.bashrc',
        line  => 'complete -C /usr/local/bin/consul consul',
        match => '^complete -C /usr/local/bin/consul consul$',
      }
```

As we're still running with Puppet 3 I can't test this with Puppet 4 (which is a req for the latest version of this module).
Would be great if somebody with a recent Puppet setup can provide a PR.

Thanks and regards

jardleex

lesinigo:
IMHO it would be better to drop a `file { '/etc/bash_completion.d/consul': }` rather than mess with root's `.bashrc`

solarkennedy:
Sounds good. I think it needs to be carefully deployed to linux-only installs first. (maybe someone from osx / windows platforms could say what the correct path is there. I think the "install" class is as good as any of a place to put this.

jardleex:
Just created the missing request. I'm open for any improvement request I can realize.

PR #581, opened 2021-09-21
added bash completion for Ubuntu and Debian - others untested

jardleex:
Hi,

this aims to solve #407.

I'm open to any improvements you'd like to have for this.
Could test it only on Ubuntu and Debian, thus set `$bash_completion` only there on true.
If the user uses a package installation (like we do) he has to adjust `$bin_dir`, depending on what package he uses. The package from HashiCorp (which we use) does not include bash completion.

Best regards

Jard


manifests/install.pp
```diff
@@ -114,4 +114,15 @@
       system => true,
     }
   }
+
+  if ($consul::bash_completion) and ($consul::install_method != 'docker' ) {
+      file { '/etc/bash_completion.d/consul':
```
bastelfreak, line 119: I'm wondering if this directory is correct? shouldn't it go into /etc/profile.d/? That works on Ubuntu/Debian/Arch/RedHat (I haven't tested others).

jardleex: Hi @bastelfreak 

thanks for the input.

The package bash-completion provides `/etc/profile.d/bash_completion.sh` which calls `/usr/share/bash-completion/bash_completion`. The later one then checks `/etc/bash_completion.d/` for any completion scripts.
The drawback I found is that `/etc/bash_completion.d/` is not created by the package bash-completion (at least version 2.8-6 which we run). Other packages do so like git, unp or what so ever.

If the completion should get realized via `/etc/profile.d/` another file content has to be used. 
How did you deploy such completions in your landcape?

bastelfreak: ah you're right. /etc/profile.d doesn't work here because `complete` is a bash builtin. the code contains stuff like `compat_dir=${BASH_COMPLETION_COMPAT_DIR:-/etc/bash_completion.d}` so I Ithink `/etc/bash_completion.d/` is indeed correct and my initial comment was just stupid. I think `/etc/bash_completion.d` should be configureable in case people have set BASH_COMPLETION_COMPAT_DIR. Those people will want to set the same path within this puppet module.

jardleex: `$bash_completion_compat_dir` has been added together with the ensure block.

If this works on other [os variants](https://github.com/solarkennedy/puppet-consul/blob/master/manifests/params.pp#L63) we may spread `$bash_completion = true`. Unfortunately I have no other distribution available to test this.

manifests/params.pp
```diff
@@ -47,6 +47,7 @@
     default: {
       # 0 instead of root because OS X uses "wheel".
       $data_dir_mode = '0755'
+      $bash_completion_compat_dir = '/etc/bash_completion.d/'
```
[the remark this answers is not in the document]
jardleex: Thanks for the feedback. I hope I got you right with this?

```
diff --git a/REFERENCE.md b/REFERENCE.md
index ffaf8db..e93022a 100644
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -222,7 +222,7 @@ Data type: `String[1]`
 
 Directory to place bash-completion file for Consul into.
 
-Default value: `$consul::params::bash_completion_compat_dir`
+Default value: `'/etc/bash_completion.d/'`
 
 ##### <a name="-consul--bin_dir"></a>`bin_dir`
 
diff --git a/manifests/init.pp b/manifests/init.pp
index 20586b2..95fb6ec 100644
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -98,7 +98,7 @@ class consul (
   String[1]                             $arch                        = $consul::params::arch,
   Optional[Stdlib::Absolutepath]        $archive_path                = undef,
   Boolean                               $bash_completion             = $consul::params::bash_completion,
-  String[1]                             $bash_completion_compat_dir  = $consul::params::bash_completion_compat_dir,
+  String[1]                             $bash_completion_compat_dir  = '/etc/bash_completion.d/',
   Stdlib::Absolutepath                  $bin_dir                     = $consul::params::bin_dir,
   Optional[String[1]]                   $binary_group                = $consul::params::binary_group,
   String[1]                             $binary_mode                 = $consul::params::binary_mode,
diff --git a/manifests/params.pp b/manifests/params.pp
index 3d3f6ed..bcabcb2 100644
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -47,7 +47,6 @@ class consul::params {
     default: {
       # 0 instead of root because OS X uses "wheel".
       $data_dir_mode = '0755'
-      $bash_completion_compat_dir = '/etc/bash_completion.d/'
       $binary_group = '0'
       $binary_mode = '0555'
       $binary_name = 'consul'
```

Issue #585, opened 2022-01-07
RFC: change dependency camp2camp/systemd module to voxpupuli/systemd


ikonia:
Consul module dependency https://forge.puppet.com/modules/camptocamp/systemd is a depreciated module that is no longer maintained - the author suggests a replacement

replacing the module with 

https://forge.puppet.com/modules/puppet/systemd - version 3.5.0 or later (it has been tested with earlier versions but setting the version on the current major / minor release seems logical) 

This module is a compatible swap out for the consul modules requirements

PR #586, opened 2022-01-07
Switch from camptocamp/systemd to puppet/systemd

bastelfreak:
The module got migrated to a new namespace

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -9,8 +9,8 @@
   "issues_url": "https://github.com/solarkennedy/puppet-consul/issues",
   "dependencies": [
     {
-      "name": "camptocamp/systemd",
-      "version_requirement": ">= 3.0.0 < 4.0.0"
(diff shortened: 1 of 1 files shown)


bastelfreak:
Fixes https://github.com/solarkennedy/puppet-consul/issues/585

metadata.json
```diff
@@ -9,8 +9,8 @@
   "issues_url": "https://github.com/solarkennedy/puppet-consul/issues",
   "dependencies": [
     {
-      "name": "camptocamp/systemd",
-      "version_requirement": ">= 3.0.0 < 4.0.0"
+      "name": "voxpupuli/systemd",
```
solarkennedy, line 12: @bastelfreak can you double check, should this be `puppet` here?

bastelfreak: yes, updated it

PR #587, opened 2022-01-19
Fix Folder Creation When They Are Not The Defaults

rachalsolutions:
This is a great module and I really like it.

One thing I do when applying this module is I specify my log folder and config folder. Because they are two levels below etc, puppet throughs an error saying it cannot create the parent directory.

This pull request fixes that by sending **log_file**, **config_dir** and **bind_dir** to a defined resource that uses **mkdir** to ensure the parent directories are present.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -113,12 +113,12 @@ class consul::config (
     }
   }
 
-  file { $consul::config_dir:
-    ensure  => 'directory',
-    owner   => $consul::config_owner_real,
-    group   => $consul::group_real,
-    purge   => $purge,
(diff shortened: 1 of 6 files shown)


solarkennedy:
Sorry, I won't accept this PR because I don't believe it is the role of this module to create the parent directories.
I recommend putting  that sort of logic in your profile_consul 
https://ospassist.puppet.com/hc/en-us/articles/360043560913-The-roles-and-profiles-pattern-What-is-it-How-do-I-use-it-to-manage-my-infrastructure-and-configuration-more-easily-

PR #590, opened 2022-04-14
Adding datacenters property for Consul policies

jonesbrennan:
I am using this puppet module with Consul version 1.11.4. On this version of Consul, maybe previous versions as well, there is a datacenters property for acl policies. I wanted to be able to set that when creating and updating policies. I have added a datacenters property to the consul_policy type and provider in this request.


types/policystruct.pp
```diff
@@ -3,6 +3,7 @@
     id            => Optional[String[1]],
     ensure        => Optional[Enum['present', 'absent']],
     description   => Optional[String[0]],
+    datacenters   => Optional[Array[String]],
```
bastelfreak, line 6: I think empty string do not make any sense, so we can prohibit it:
```suggestion
    datacenters   => Optional[Array[String[1]]],
```

jonesbrennan:
I've updated the test cases to account for the new datacenters parameter and ran the tests locally to ensure they pass.

solarkennedy:
Thanks!

solarkennedy:
Thanks!

PR #605, opened 2022-10-23
Package install: Fix ordering when data_dir isnt managed

bastelfreak:
Previously, we ensured that the package resources comes before creating
a file resource for the data dir. This doesn't always make sense. The
rpm/deb packages (depending in the platform) will create the directory for us, so we
don't need to care about it.

Contains https://github.com/voxpupuli/puppet-consul/pull/604 to proove it's a bug.


bastelfreak:
Implemented in #610

PR #609, opened 2022-10-24
Drop legacy init scripts

bastelfreak:
We don't support those sysv init scripts since a long time, we can delete them. We dropped support for the related Operating Systems in the past.


bastelfreak:
My feelings tell me this is just an enhancement because we cleanup some leftovers. it doesn't need to be a breaking change because we dropped support for those legacy OSes in the past already.

spec/classes/init_spec.rb
```diff
@@ -725,12 +593,7 @@
...
-        if facts[:os]['release']['major'].to_i == 6
-          context 'On a redhat 6 based OS' do
-            it { is_expected.to contain_class('consul').with_init_style('redhat') }
-            it { is_expected.to contain_file('/etc/init.d/consul').with_content(%r{daemon --user=consul}) }
-          end
-        elsif facts[:os]['release']['full'] == '2016.09'
+        if facts[:os]['release']['full'] == '2016.09'
           context 'On an Amazon based OS' do
             it { is_expected.to contain_class('consul').with_init_style('redhat') }
             it { is_expected.to contain_file('/etc/init.d/consul').with_content(%r{daemon --user=consul}) }
```
evgeni, line 599: I'd expect these to also fail now?

evgeni:
> My feelings tell me this is just an enhancement because we cleanup some leftovers. it doesn't need to be a breaking change because we dropped support for those legacy OSes in the past already.

*Technically* it's still possible to run Debian 10 (and maybe Debian 11) with sysvinit, so *technically* this is a breaking change.

Whether it was ever *supported* by us, is a completely different question.

spec/classes/init_spec.rb
evgeni: But I guess they are not even executed, as that is not listed as `supported os`?

spec/classes/init_spec.rb
```diff
@@ -740,43 +603,25 @@
...
-        elsif facts[:os]['release']['major'].to_i == 7
-          context 'On a redhat 7 based OS' do
+        elsif facts[:os]['release']['major'].to_i >= 7
+          context 'On a RedHat 7 and newer based OS' do
             it { is_expected.to contain_class('consul').with_init_style('systemd') }
             it { is_expected.to contain_file('/etc/systemd/system/consul.service').with_content(%r{consul agent}) }
           end
-        elsif facts[:os]['release']['major'].to_i == 20
-          context 'On a fedora 20 based OS' do
+        elsif facts[:os]['release']['major'].to_i >= 20
```
evgeni, line 611: Given the `>= 7` above, this will never match? (the tests are identical, so there is no issue as such, just weird code)

bastelfreak: fixed

spec/classes/init_spec.rb
bastelfreak: Ah, because we don't support that OS and it's not listed in the metadata.json, so the OS is never set to 2016.19. I cleaned it up.

evgeni (approved):
ACK on the change
no idea about enhancement vs breaking :)

PR #612, opened 2022-10-26
manage_repo: ensure that we refresh the package list before installing consul

bastelfreak:
by including the hashistack_repo class, we get the correct apt repo. It will also notify the apt_update exec resource. We just need to ensure that the exec happens before we install consul.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -65,7 +65,11 @@ class consul::install {
     'package': {
       if $consul::manage_repo {
         include hashi_stack::repo
-        Class['hashi_stack::repo'] -> Package[$consul::package_name]
+        if $facts['os']['family'] == 'Debian' {
+          Exec['apt_update'] -> Package[$consul::package_name]
+        } else {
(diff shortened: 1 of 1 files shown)


kenyon (commented):
Should it be `Class['apt::update']` instead of the Exec?

bastelfreak:
https://github.com/puppetlabs/puppetlabs-apt#update-the-list-of-packages explicitly mentions the exec that should be notified/depends on.

attachmentgenie (approved):

kenyon:
@bastelfreak ah I read that section as a description of how the module works, and considered the Exec to be an implementation detail. I was looking at this section: https://github.com/puppetlabs/puppetlabs-apt#adding-new-sources-or-ppas

It's confusing at the least. I haven't been able to get apt updates to happen in the same puppet run as a package installation, consistently.

bastelfreak:
It was definetly broken in many older puppetlabs-apt releases, but with the depencendy on the exec it's working fine for me since some time.

PR #614, opened 2022-10-27
Implement Sensitive support for config

bastelfreak:

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -111,6 +111,6 @@ class consul::config (
     owner   => $consul::config_owner_real,
     group   => $consul::group_real,
     mode    => $consul::config_mode,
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -3,7 +3,6 @@
 # This class is meant to be called from consul
 # It sets variables according to platform
 #
-# @api private
```
ekohl, line 6: Accidental removal?

bastelfreak: no. the params.pp doesn't have assert_private() (and adding it will break it). That comment was added by accident. I cleaned it up.

ekohl: It still is a private class as users aren't supposed to call it AFAIK. I'd think it should stay so `puppet-strings` doesn't document it.

bastelfreak: fair point. Added it back.

sebastianrakel (approved):

PR #615, opened 2022-10-27
systemd service: Switch erb to epp template

bastelfreak:


templates/consul.systemd.epp
```diff
@@ -0,0 +1,37 @@
...
+      Boolean $enable_beta_ui
+| -%>
+# THIS FILE IS MANAGED BY PUPPET
+[Unit]
+Description=Consul Agent
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+<% if $config_hash['retry_join'] == undef or $config_hash['retry_join'].length < 2 { -%>
```
bastelfreak, line 17: I checked https://github.com/voxpupuli/puppet-consul/pull/552 and I'm not sure if this is really required. But we had the logic in the erb template, so I just ported it.

templates/consul.systemd.epp
```diff
@@ -0,0 +1,37 @@
...
+Type=exec
+<% } else { -%>
+Type=notify
+<% } -%>
+<% if $allow_binding_to_root_ports == true { -%>
```
ekohl, line 22: It's a boolean so this is safe too:
```suggestion
<% if $allow_binding_to_root_ports { -%>
```

templates/consul.systemd.epp
```diff
@@ -0,0 +1,37 @@
...
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+<% } -%>
+<% if $enable_beta_ui == true { -%>
```
ekohl, line 25:
```suggestion
<% if $enable_beta_ui { -%>
```

templates/consul.systemd.epp
ekohl: If you're changing it, I'd keep the `config_hash` of the template and move it into the Puppet code. Introduce a simple `String[1] $service_type` (or `Enum`) parameter. Mostly since eventually I'd consider using [`systemd::manage_unit`](https://github.com/voxpupuli/puppet-systemd/blob/master/manifests/manage_unit.pp) once that's out in a release.

bastelfreak:
@ekohl I updated it as suggested

PR #616, opened 2022-10-27
mark internal classes as private

bastelfreak:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -3,6 +3,7 @@
 # This class is meant to be called from consul
 # It sets variables according to platform
 #
+# @api private
 class consul::params {
   $manage_repo = false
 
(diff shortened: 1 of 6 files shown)


manifests/params.pp
```diff
@@ -3,6 +3,7 @@
 # This class is meant to be called from consul
 # It sets variables according to platform
 #
+# @api private
```
rwaffen, line 6: no assert here?

manifests/windows_service.pp
```diff
@@ -8,11 +8,12 @@
 #
 # [*service_name*]
```
ekohl, line 9: Looks like this is using an old style to document parameters. Will that conflict?

ekohl (approved):

manifests/params.pp
bastelfreak: ah that comment was a leftover. we cannot mark the params class as private. I removed the comment.

PR #617, opened 2022-10-31
Convert classes to puppet-strings

bastelfreak:


manifests/check.pp
```diff
@@ -1,62 +1,25 @@
...
-#   Value in seconds for the interval between runs of the check
-#
-# [*notes*]
-#   Human readable description of the check
-#
-# [*script*]
+# @summary Sets up a Consul healthcheck
+# @see http://www.consul.io/docs/agent/checks.html
+#
+# @param ensure Define availability of check. Use 'absent' to remove existing checks. Defaults to 'present'
```
alexjfisher, line 4: Do you not prefer the style where the description is on the line below?  (Also you can leave out `Defaults to 'present'` here as puppet-strings will produce that information itself).

manifests/watch.pp
```diff
@@ -1,51 +1,20 @@
...
-# [*state*]
-#   Watch a state change on a service healthcheck.
-#
-# [*token*]
-#   String to override the default token.
-#
-# [*type*]
-#   Type of data to watch. (Like key, service, services, nodes)
+# @summary Sets up Consul watch, to span commands when data changes.
+# @see http://www.consul.io/docs/agent/watches.html
```
alexjfisher, line 3: From https://puppet.com/docs/puppet/7/puppet_strings_style.html#puppet_strings_style-strings-style-comment

> Separate comment sections (such as @summary, @example, or the @param list) with a blank comment line (that is, a # with no additional content), to improve readability.

bastelfreak: fixed

manifests/check.pp
bastelfreak: I removed the defaults mentions. My personal preference is to have it in a single line, and my guess is that we've a slight majority of modules with that style. But I can live with multiline as well.

bastelfreak: maybe that's something for our styleguide :thinking:

alexjfisher: My preference is the opposite :)

The upstream styleguide says use either, but presumably the author preferred the indented description style as that's used in all the examples.

sebastianrakel (approved):

PR #637, opened 2023-08-21
Update default consul version 1.2.3->1.16.1

bastelfreak:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -137,7 +137,7 @@ class consul (
   Enum['stopped', 'running']            $service_ensure              = 'running',
   Hash                                  $services                    = {},
   String[1]                             $user                        = $consul::params::user,
-  String[1]                             $version                     = '1.2.3',
+  String[1]                             $version                     = '1.16.1',
   Hash                                  $watches                     = {},
   Optional[String[1]]                   $shell                       = $consul::params::shell,
   Boolean                               $enable_beta_ui              = false,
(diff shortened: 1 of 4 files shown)


spec/acceptance/class_spec.rb
```diff
@@ -117,51 +117,7 @@ class { 'consul':
     end
 
     describe command('consul version') do
-      its(:stdout) { is_expected.to match %r{Consul v1.2.0} }
```
bastelfreak, line 120: the whole block was a duplicate, we don't need to test it multiple times.

manifests/init.pp
```diff
@@ -137,7 +137,7 @@
   Enum['stopped', 'running']            $service_ensure              = 'running',
   Hash                                  $services                    = {},
   String[1]                             $user                        = $consul::params::user,
-  String[1]                             $version                     = '1.2.3',
+  String[1]                             $version                     = '1.16.1',
```
smortex, line 140: I generally wonder if it is not better to _not_ set any default version, and mandate the user to explicitly put what they want.  This mean that we have to set an arbitrary version for testing, but we do not have to live with having to choose to default to an old-legacy-deprecated version after a few years or break backwards compatibility when a new version is available.

smortex (commented):
See in-line comment, but if not appropriate with this module (I am not a user of consul) I am fine with this.

manifests/init.pp
bastelfreak: I considered that and I don't know if I like it or not. We need to do a major release anyway so I thought it's a good time to bump the version, also consul supports upgrades between versions.

smortex (approved):

PR #638, opened 2023-08-21
CI: Use Type=exec instead of Type=notify

bastelfreak:
Hashicorp changed this some time back. Type=notify doesn't work properly in docker containers and breaks our CI.


hashworks (approved):
LGTM

PR #642, opened 2023-10-28
Release 8.0.0

mj:

This PR prepare the release of puppet-consul 8.0.0. I've opted for a new major version because Puppet 6 support has been dropped.


smortex (commented):
LGTM!  I am not a user of consul so would like another review before we kick a release.  Also we may want to include #637 in this major release?  cc @bastelfreak

smortex (approved):

PR #643, opened 2023-10-31
Update default version 1.16.1->1.16.3

bastelfreak:


bastelfreak:
we need to do a major release anyways so it makes sense to bump the version now.

PR #655, opened 2024-02-14
use X-Consul-Token header instead of token query parameter

jardleex:
Hi,

this change replaces the usage of the token query paramter with the HTTP header X-Consul-Token when interacting with ACLs, key/values and prepared queries.
The header was introduced in Consul [0.6.0](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#060-december-3-2015).
The change is necessary as the token query parameter is marked [deprecated in Consul 1.15.x](https://developer.hashicorp.com/consul/docs/upgrading/upgrade-specific#deprecating-authentication-via-token-query-parameter). Using it triggers deprecation warnings in the Consul logs each time this module runs on a server.

~~~bash
Jan 26 10:44:11 consul03 consul[11454]: agent.http: This request used the token query parameter which is deprecated and will be removed in Consul 1.17: logUrl=/v1/query?token=<hidden>
~~~

Luckily it was not removed yet so there is still time to fix it.
I've tested the changes with our internal Puppet module in Vagrant.

This pull request was created by my best knowladge. I'm working as system administrator and have no ruby expertise but got some help from colleagues in our company. Thus said I'm open for any suggestion to improve this.

Thanks in advance!

#### This Pull Request (PR) fixes the following issues
n/a.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1110,7 +1110,7 @@ The following parameters are available in the `consul_acl` type.
 
 Token for accessing the ACL API
 
-Default value: `''`
+Default value: `anonymous`
 
 ##### <a name="-consul_acl--api_tries"></a>`api_tries`
 
@@ -1179,7 +1179,7 @@ The following parameters are available in the `consul_key_value` type.
 
 Token for accessing the ACL API
 
-Default value: `''`
+Default value: `anonymous`
 
 ##### <a name="-consul_key_value--api_tries"></a>`api_tries`
 
@@ -1373,7 +1373,7 @@ The following parameters are available in the `consul_prepared_query` type.
 
 Token for accessing the ACL API
 
-Default value: `''`
+Default value: `anonymous`
 
 ##### <a name="-consul_prepared_query--api_tries"></a>`api_tries`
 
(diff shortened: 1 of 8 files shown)


zilchms:
try running
```
bundle install
bundle exec rake strings:generate:reference
```
that should fix the test and get the CI running :)

bastelfreak:
@jardleex thanks for the PR. We run some acceptance tests in https://github.com/voxpupuli/puppet-consul/blob/master/spec/acceptance/class_spec.rb. Do we need to adjust them for other versions? Does your patch work on older  consul versions as well?

jardleex:
We only tested it on Consul 1.17.0.
According to the official [changelog](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#060-december-3-2015) the header support was added in Consul 0.6.0 in 2015 with this [PR](https://github.com/hashicorp/consul/blob/main/CHANGELOG.md#060-december-3-2015).
Currently the Puppet module version > 4.0.0 requires at least Consul 1.1.0 from May 2018 which supports the X-Consul-Token header already.

PR #656, opened 2024-03-07
Allow to install consul through local archive files

2fa:
You should be allowed to install consul through local files besides http URIs (similar to how Nomad module doing it right now)


TheMeier:
Please update `REFERENCE.md`
can be done via: `bundle exec rake strings:generate:reference`

PR #714, opened 2026-09-03
Allow specifying custom reload command

### LINKED ISSUE #629: alternative commands to Consul Reload
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7
- Module version: HEAD

## What are you seeing

The Exec inside `consul::reload_service` never works for me.  Trying to reload Consul via the Consul CLI is the most universal method, however it means a token must be in place for it to work.  I use ACLs, and I will not pass any token to Puppet for security reasons.

If I were to submit a PR to optionally change the behaviour of `consul::reload_service` to call out to `systemctl reload consul` if `$facts['systemd']` plus another class Boolean, would that be acceptable? It's not pretty, but it's the best I've got.

The other thing I can do is spaceship operator and override:

```
Exec <| title == 'reload consul service' |> {
  command => 'systemctl reload consul'
}
```

Which is even nastier, but it'll work for me.

## Output log

```
Notice: /Stage[main]/Consul::Reload_service/Exec[reload consul service]/returns: Error reloading: Unexpected response code: 403 (Permission denied: anonymous token lacks permission 'agent:write' on "hostname". The anonymous token is used implicitly when a request does not specify a token.)
Error: /Stage[main]/Consul::Reload_service/Exec[reload consul service]: Failed to call refresh: 'consul reload -http-addr=127.0.0.1:8500' returned 1 instead of one of [0]
Error: /Stage[main]/Consul::Reload_service/Exec[reload consul service]: 'consul reload -http-addr=127.0.0.1:8500' returned 1 instead of one of [0]
```

hbro:
Ran into the same problem as lukebigum encountered in issue #629. Did not see a related PR so figured I'd write a potential fix to the problem myself.

Basically allow setting a custom reload command. If it's defined, use that. If not, use the existing logic.

#### This Pull Request (PR) fixes the following issues
Fixes #629


rwaffen (approved):
LGTM

